PR #11, opened 2025-01-20
[blobfish465] iP

blobfish465:
### Blob Task Manager 

Blob is:

- Text-based
- User-friendly
- ~~Fast~~ _SUPER_ FAST to use 

Here's how to get started:

1. Clone the repository from [here](https://github.com/blobfish465/ip)
2. Compile the Java files.
3. Run the application and start managing your tasks!
4. Add your task
5. Let it manage your task for you! :smile:

And it is **FREE!**

Features:

- [X] Managing tasks: Add, delete, and mark tasks as done.
- [X] Time-bound tasks: Set deadlines and events with specific times. 
- [X] Search: Quickly find tasks by keywords.
- [ ] Reminders (Coming Soon!)

```java
    new Blob("data/Blob.txt").run();
}
```

Enjoy managing your tasks with Blob! :memo:


src/main/java/blob/Main.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.IOException;
+
+import blob.controller.MainWindow;
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+
```
yuqiannemo, line 10: The ordering of import statements LGTM.

src/main/java/blob/command/Command.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException;
+
+/**
+ * Represents a command that can be executed in the Blob application.
+ * Each command performs an action on the provided {@link TaskList}, updates the user interface
+ * through {@link Ui}, and may interact with {@link Storage} to persist changes.
+ */
+public interface Command {
+
```
yuqiannemo, line 15: Should this line break (line 16) be removed?

src/main/java/blob/model/Task.java
```diff
@@ -0,0 +1,116 @@
...
+            System.out.println("Skipping invalid or incomplete line: " + line);
+            return null;
+        }
+
+        String type = parts[0].trim();
+        boolean isDone = parts[1].trim().equals("1");
+        String description = parts[2].trim();
+        Task task = null;
+
+        switch (type) {
```
yuqiannemo, line 92: The coding standard for "switch" statements are closely followed!

src/main/java/blob/command/AddCommand.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Executes the add command which involves adding a task to the task list
+     * and updating the UI to show the added task.
+     *
+     * @param tasks The task list to which the task will be added.
+     * @param ui The UI to interact with the user.
+     * @param storage The storage used to save the updated task list.
+     */
```
yuqiannemo, line 32: Nice JavaDoc added! Also, when writing Javadocs for overridden methods, the **@inheritDoc tag** can be used to reuse the header comment from the parent method but with further modifications e.g., when the method has a slightly different behavior from the parent method (this applies to other subclasses of Command class too).

src/main/java/blob/TaskList.java
```diff
@@ -0,0 +1,95 @@
+package blob;
```
yuqiannemo, line 1: The naming LGTM.

weiwong834 (commented):
Code looks good! No naming issues found

PR #13, opened 2025-01-20
[ChinZJ] iP

ChinZJ:
# Botling  
> "Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live." - John Woods [(source)](https://www.goodreads.com/quotes/248194-always-code-as-if-the-guy-who-ends-up-maintaining)  
  
Botling is a task list application that is     
* text-based
* easy to learn
* easy to use  
  
All you need to do is,
1. download it from [here](https://github.com/ChinZJ/ip/releases/tag/A-Jar).  
1. double-click it.
1. add your tasks.  
1. let it manage your tasks for you. 🤙   
  
And it is *FREE*!  
  
Features:  
- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders
- [ ] User friendly Graphical user interface  
  
---  
```
package botling;

import botling.commands.CommandParser;
import java.util.Scanner;

/**
 * Main class where application starts.
 */
public class Botling {
    /**
     * Generates objects required for program to function.
     * Handles the main loop of the program
     */
        Scanner reader = new Scanner(System.in);
        TaskList tasks = new TaskList();
        CommandParser cmdParse = new CommandParser();

        // Generates the startup.
        System.out.println(cmdParse.start(tasks));

        // User input loop.
        while (tasks.isOpen()) {
            // Leveraging on object pass by reference
            System.out.println(cmdParse.parse(reader.nextLine(), tasks));
        }
        reader.close();
    }
}
```


src/main/java/botling/DateParser.java
```diff
@@ -0,0 +1,63 @@
...
+        tempMap.put("dd/MM/yyyy", DateTimeFormatter.ofPattern("dd/MM/yyyy"));
+        tempMap.put("dd/MM/yyyy HHmm", DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm"));
+        tempMap.put("dd/MM/yy", DateTimeFormatter.ofPattern("dd/MM/yy"));
+        tempMap.put("dd/MM/yy HHmm", DateTimeFormatter.ofPattern("dd/MM/yy HHmm"));
+        tempMap.put("dd MMM yyyy", DateTimeFormatter.ofPattern("dd MMM yyyy"));
+        tempMap.put("dd MMM yyyy HHmm", DateTimeFormatter.ofPattern("dd MMM yyyy HHmm"));
+        tempMap.put("dd MMM yy", DateTimeFormatter.ofPattern("dd MMM yy"));
+        tempMap.put("dd MMM yy HHmm", DateTimeFormatter.ofPattern("dd MMM yy HHmm"));
+        DATE_MAPPER = Collections.unmodifiableMap(tempMap);
+    }
```
gengyudong, line 33: I really like this part where you took into account different DateTime formats!

src/main/java/botling/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Stores all <code>Task</code> objects.
+ * Possesses core list related functionality required to interact with <code>Task</code> objects.
+ */
+public class TaskList {
+    private static final boolean TASK_IS_DONE = true;
+
+    private ArrayList<Task> tasks;
+    private boolean isActive;
```
gengyudong, line 16: Would it be more efficient to initialize the 'isActive' field to be true as compared to declaring the 'TASK_IS_DONE' variable and then assigning it to the 'isActive' field in the constructor?

src/main/java/botling/DateParser.java
```diff
@@ -0,0 +1,63 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Check if user input is valid as a date object.
+ */
+public class DateParser {
```
gengyudong, line 15: Good naming convention - setting class name to be in verb form!

src/main/java/botling/commands/commandtypes/DeadlineCmd.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Parses deadline commands.
+ */
+public class DeadlineCmd implements TasksCmd {
+    private static final int IDX_NAME = 0;
+    private static final int IDX_BY = 1;
+
+    /**
+     * Creates a deadline task.
+     */
```
gengyudong, line 28: Would it be clearer to include the params of this method in the Javadoc comments?

gengyudong (commented):
Overall, good job in following the coding standards! Your coding skills are really good! Keep up the great work!

src/main/java/botling/TaskList.java
ChinZJ: Not too sure if setting true without that would be considered a magic number of sorts? What do you think.

src/main/java/botling/commands/commandtypes/DeadlineCmd.java
ChinZJ: Ah yes, forgot about it completely, thank you for the reminder.

src/main/java/botling/DateParser.java
ChinZJ: Thank you!

src/main/java/botling/DateParser.java
ChinZJ: Thank you!

PR #28, opened 2025-01-22
[xumarcus] iP

xumarcus:
# Zoe
Features:
- text-based
- easy to learn

Supported tasks:
1. Todo
2. Deadline
3. Event

```java
try (Reader reader = new FileReader(path)) {
    List<Task> tasks = gson.fromJson(reader, new TypeToken<List<Task>>() {
    }.getType());
    if (tasks == null) throw new ZoeIOException("File is empty");
    return tasks;
}
```
Coming soon:
- [ ] Reminders

:grinning:
> Trivia: In leetspeak, 2103 corresponds to z1oe.

Zoe uses [gson](https://github.com/google/gson), like `gson.toJson(taskList, writer)`, to **store** and *load* tasks, ~~i.e.~~.


src/main/java/zoe/adapters/gson/LocalDateTypeAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+import java.lang.reflect.Type;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+// https://mkyong.com/java/gson-supports-java-8-date-time-types/#default-gson-does-not-support-java-8-date-time-type
+public class LocalDateTypeAdapter implements JsonSerializer<LocalDate>, JsonDeserializer<LocalDate> {
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
varuuuun, line 32: `formatter` isnt in capital case, should it be in SCREAMING_SNAKE_CASE?

src/main/java/zoe/ZoeException.java
```diff
@@ -0,0 +1,7 @@
+package zoe;
+
+public class ZoeException extends Exception {
```
varuuuun, line 20: What is this error for? Some javadoc here would be nice. Especially since you have 2 error classes, to differentiate the use cases.

src/main/java/zoe/ZoeIOException.java
```diff
@@ -0,0 +1,7 @@
+package zoe;
+
+public class ZoeIOException extends ZoeException {
```
varuuuun: What is this error for? Some javadoc here would be nice. Especially since you have 2 error classes, to differentiate the use cases.

src/main/java/zoe/Zoe.java
```diff
@@ -0,0 +1,81 @@
...
+import javafx.application.Platform;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+import zoe.command.Command;
+import zoe.command.Parser;
+import zoe.storage.Storage;
+import zoe.task.TaskList;
+import zoe.ui.Ui;
```
varuuuun, line 15: The order of your imports does not follow the required standards:
image

src/main/java/zoe/storage/Storage.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.Writer;
+import java.util.List;
+
+@AllArgsConstructor
+public class Storage {
+    // Call registerTypeAdapter(LocalDate.class, new zoe.adapters.gson.LocalDateTypeAdapter()) in zoe.adapters.gson.TaskTypeAdapter!
+    private final static Gson gson = new GsonBuilder()
```
varuuuun: Should this variable name be in all caps, to align with the SCREAMING_SNAKE_CASE requirement?

varuuuun (commented):
Really interesting work, some minor nits found.  

Generally your import statements across all files do not seem to follow the coding standard, refer to the comment I left under Zoe.java file. You might want to improve on that.

src/main/java/zoe/Zoe.java
xumarcus: Fixed

src/main/java/zoe/storage/Storage.java
xumarcus: This has side effect, so probably not

src/main/java/zoe/adapters/gson/LocalDateTypeAdapter.java
xumarcus: Need to clarify. Most code samples use lower case

src/main/java/zoe/storage/Storage.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return a list of tasks loaded from the file
+     * @throws ZoeIoException if the file is not found, cannot be read, or if parsing fails
+     */
+    public List<Task> load() throws ZoeIoException {
+        try (Reader reader = new FileReader(path)) {
+            List<Task> tasks = gson.fromJson(reader, new TypeToken<List<Task>>() {
+            }.getType());
+            if (tasks == null) {
+                throw new ZoeIoException("File is empty");
+            }
```
sreelakshmiharidos, line 83: Perhaps you don't need to throw an exception for empty file, since it's a valid state. Instead, maybe just return an empty list and hence can reduce complexity in try-catch block.

src/main/java/zoe/ui/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+ */
+public class Ui extends AnchorPane {
+    @FXML
+    private ScrollPane scrollPane;
+    @FXML
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    @SuppressWarnings("unused")
```
sreelakshmiharidos, line 32: Perhaps add a comment with @SuppressWarnings to convey rationale and improve readability?

src/main/java/zoe/ui/Ui.java
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @param tasks The list of tasks to be displayed.
+     */
+    public void showList(List<? extends Task> tasks) {
+        reply("Here are the matching tasks in your list:\n"
+                + IntStream
+                .range(0, tasks.size())
+                .mapToObj(i -> String.format("%d. %s", i + 1, tasks.get(i).getFullDescription()))
+                .collect(Collectors.joining("\n")));
+    }
```
sreelakshmiharidos, line 141: Maybe use multiline string literal instead of concatenation to have consistent formatting with the rest of the code, thereby making the code easier to read and maintain?
```suggestion
    public void showList(List<? extends Task> tasks) {
    String taskListStr = IntStream
            .range(0, tasks.size())
            .mapToObj(i -> String.format("%d. %s", i + 1, tasks.get(i).getFullDescription()))
            .collect(Collectors.joining("\n"));

    reply(String.format("""
            Here are the matching tasks in your list:
            %s""", taskListStr));
}
```

src/main/java/zoe/command/MarkCommand.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+@Getter
+@AllArgsConstructor
+public class MarkCommand implements Command {
+    public static final String COMMAND_WORD = "mark";
+    private final int index;
+
+    @Override
+    public void execute(Storage storage, TaskList taskList, Ui ui) throws ZoeException {
+        assert(0 <= index && index < taskList.size());
```
sreelakshmiharidos: Perhaps use explicit check instead of assert as assert is meant for debugging and not runtime validation?
```suggestion
        if (index < 0 || index >= taskList.size()) {
    throw new ZoeException("Invalid task index: " + index);
}
```

sreelakshmiharidos (commented):
LGTM. Just a few nits to improve code quality even more!

PR #33, opened 2025-01-23
[clx3210] iP

clx3210:
This bot frees your mind of clutter by managing your tasks. It's:

- text-based
- easy to learn
- ~~FAST~~ *SUPER* FAST to use

All you need to do is,

1. download it from [here](https://github.com/clx3210/ip/releases/tag/A-Jar).
1. double-click it.
1. add your tasks.
1. let it manage your tasks for you :wink:

And it's **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you are a Java programmer, you can practice it too. Here's the `main` method:
```java
public class Main {
    }
}
```

Contribute by cloning the repo (`git clone https://github.com/clx3210/ip.git`).


src/main/java/olivero/Olivero.java
```diff
@@ -0,0 +1,70 @@
...
+import olivero.exceptions.CommandExecutionException;
+import olivero.exceptions.CommandParseException;
+import olivero.exceptions.StorageLoadException;
+import olivero.parsers.Parser;
+import olivero.storage.Storage;
+import olivero.tasks.TaskList;
+
+/**
+ * Represents the main entry point for the program.
+ */
```
iuhiah, line 16: All the JavaDocs are descriptive and easy to understand :)

src/main/java/olivero/common/Responses.java
```diff
@@ -0,0 +1,48 @@
+package olivero.common;
+
+/**
+ * Common display responses that will be shown to the user.
+ */
+public final class Responses {
```
iuhiah, line 6: Really like this idea of having a fixed set of responses, might implement this in my own version :>

src/main/java/olivero/tasks/Task.java
```diff
@@ -0,0 +1,65 @@
...
+    public void setDone(boolean done) {
+        isDone = done;
+    }
+
+    /**
+     * Returns true if this task is completed and false otherwise.
+     *
+     * @return Task completion status.
+     */
+    public boolean isDone() {
```
iuhiah, line 37:
```suggestion
    public boolean getIsDone() {
```

Don't think this follows getter naming conventions, should be getIsDone()!

src/main/java/olivero/tasks/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toFormattedString() {
+        int doneStatus = isDone() ? 1 : 0;
+        String dateString = DateParser.asInputDateString(endDate);
+        return "D" + " | " + doneStatus + " | " + getDescription() + " | " + dateString;
```
iuhiah:
```suggestion
        return "D" + " | " + super() + " | " + dateString;
```

I would recommend extracting `doneStatus + " | " + getDescription()` in the subclasses to the Task superclass and calling super() instead to maintain the abstraction barrier (more so for the isDone attribute since it does not seem to be accessed anywhere else)

src/test/java/olivero/parsers/ParserTest.java
```diff
@@ -0,0 +1,232 @@
...
+ */
+public class ParserTest {
+
+    private static final String MESSAGE_EXPECTED_INVALID_INTEGER = "Did you pass "
+            + "in a valid integer? Your input: %s";
+
+    @Test
+    public void parse_blankArguments_exceptionThrown() {
+        String expected = "W-WHAT?! "
+                + "I do not understand what you just said :(";
```
iuhiah, line 34:
```suggestion
        String expected = Responses.RESPONSE_UNKNOWN_COMMAND
```

Is there any reason the expected responses are not imported from common.Responses? It would help if you had to make changes/add to the responses along the way (less lines to edit)

src/main/java/olivero/parsers/TaskParser.java
```diff
@@ -0,0 +1,92 @@
...
+     * @throws TaskParseException If the provided format of the raw tasks is invalid during parsing.
+     */
+    public TaskList parse(String rawTasks) throws TaskParseException {
+        TaskList taskList = new TaskList();
+        if (rawTasks.isEmpty()) {
+            return taskList;
+        }
+        String[] lines = rawTasks.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] args = line.split(Pattern.quote(SEPARATOR));
```
iuhiah:
```suggestion
            String[] args = line.split(Pattern.quote(SEPARATOR), limit);
```

Consider taking into account cases where the task description has '|' with split()'s limit argument

src/main/java/olivero/parsers/Parser.java
```diff
@@ -0,0 +1,196 @@
...
+import olivero.exceptions.CommandParseException;
+import olivero.exceptions.UnsupportedCommandException;
+import olivero.tasks.Deadline;
+import olivero.tasks.Event;
+import olivero.tasks.ToDo;
+
+/**
+ * Parses commands from the user input and returns a {@code Command} object.
+ */
+public class Parser {
```
iuhiah, line 26:
```suggestion
public class InputParser {
```

Maybe renaming this to InputParser (or something similar) would better follow the naming of the other parsers, I assumed that this was the superclass of DateParser and TaskParser

iuhiah (commented):
Code is relatively easy to read and understand, do remember to stick to the style conventions (remove all the unnecessary linebreaks)! Class abstraction feels slightly overwhelming, I would think about extracting common methods to existing classes (e.g. the methods in CommandResult could be absorbed into the main Command class and have the subclasses override the method as needed).

Nit: TaskList tests are too long IMO, particularly the methods that test the adding of a lot of tasks. Testing adding one of each type of task would be sufficient if the test goal was to add multiple tasks as a single string input.

src/main/java/olivero/tasks/Task.java
clx3210: Hi, since this is a boolean method, it has been named to sound like a boolean based on the coding conventions

src/main/java/olivero/Olivero.java
clx3210: Thanks!

src/test/java/olivero/parsers/ParserTest.java
clx3210: This helps to keep the testing code independent of the source code, to prevent false negatives

src/main/java/olivero/parsers/DateParser.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Returns the string representation of the provided {@code LocalDateTime} object
+     * formatted using {@link #DISPLAY_DATE_FORMATTER}.
+     *
+     * @param date The {@code LocalDateTime} object to be formatted into a string.
+     * @return The formatted string of the given date.
+     * @throws DateTimeException If an error occurs during formatting
+     */
+    public static String asDisplayDateString(LocalDateTime date) {
```
niha81: Perhaps the method could be named something like 'formatForDisplay'?

src/main/java/olivero/parsers/DateParser.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Returns the string representation of the provided {@code LocalDateTime} object
+     * formatted using {@link #INPUT_DATE_FORMATTER}.
+     *
+     * @param date The {@code LocalDateTime} object to be formatted into a string.
+     * @return The formatted string of the given date.
+     * @throws DateTimeException If an error occurs during formatting
+     */
+    public static String asInputDateString(LocalDateTime date) {
```
niha81: Similar to the method above maybe the name could be changed to 'formatForInput'

niha81 (commented):
LGTM! I like that your method and variable names are very clear it really helps with understanding the code : )

PR #37, opened 2025-01-23
[WhyAre] iP

WhyAre:
# Heading

- bullet list

1. numbered list

```java
System.out.println("I use vim btw");
```

- [ ] task list

:sob: 

> block quote

[Click here to install neovim](https://neovim.io)

`inline code`

I use **vim** (~instead of emacs~) _btw_


src/main/java/jank/JankBot.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+
+    /**
+     * Performs the action that's supplied into the function
+     *
+     * @param line command as a String[]
+     * @throws JankBotException
+     */
+    static void processCommand(String[] line) throws JankBotException {
```
deborahluyx: Should this function name be renamed to provide more explanation on how the command is processed? Perhaps `executeUserCommand`?

src/main/java/jank/Storage.java
```diff
@@ -0,0 +1,51 @@
...
+public class Storage {
+    /**
+     * Stores the tasks into a file
+     *
+     * @param filepath relative path to the source file
+     * @return List of tasks as a TaskList object
+     */
+    static TaskList loadTasks(String filepath) {
+        try (var fin = new FileInputStream(filepath); var ois = new ObjectInputStream(fin)) {
+            if (ois.readObject() instanceof TaskList tasklist) {
```
deborahluyx, line 25: Perhaps use camel case for this?

deborahluyx (commented):
Looks really good overall! Not much comments left because the code is good 👍🏻 Keep it up!

src/main/java/jank/TaskImpl.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param line  Input line
+     * @param flags array of mandatory flags
+     * @return map of flag to the value associated to the flag
+     * @throws JankBotException Throws an error when not all the flags are present in the input string
+     */
+    static Map<String, String> parseFlags(String[] line, String[] flags) throws JankBotException {
+        var map = Arrays.stream(flags)
+                        .collect(Collectors.toMap(Function.identity(), x -> new StringBuffer(), (left, right) -> right,
+                                HashMap<String, StringBuffer>::new));
+        map.put("", new StringBuffer());
```
hzk-lab: I think you may consider adding a brief comment explaining the purpose of map.put("", new StringBuffer());. Right now, it's not immediately clear why an empty key is needed.

src/main/java/jank/Parser.java
```diff
@@ -0,0 +1,102 @@
+package jank;
```
hzk-lab: That's good use of Java Stream here!

src/main/java/jank/TaskList.java
```diff
@@ -0,0 +1,71 @@
...
+    TaskList(List<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    Task add(Task task) {
+        tasks.add(task);
+        return task;
+    }
+
+    Task remove(int index) {
```
hzk-lab: I think you can consider handling IndexOutOfBoundsException in remove(int index) to prevent unexpected crashes if the index is invalid.

src/main/java/jank/TaskList.java
```diff
@@ -0,0 +1,71 @@
...
+        return task;
+    }
+
+    /**
+     * Returns a list of task which has the query in the title
+     *
+     * @param query Query string
+     * @return list of tasks
+     */
+    List<Task> find(String query) {
```
hzk-lab: What happens if query is null? Maybe add a null check before filtering to prevent NullPointerException.

src/main/java/jank/Parser.java
```diff
@@ -0,0 +1,102 @@
...
+        return new DeleteCommand(index);
+    }
+}
+
+record MarkCommand(int index, boolean isMarked) implements Command {
+    static MarkCommand parse(String[] line) throws JankBotException {
+        boolean isMarked = line[0].equalsIgnoreCase("mark");
+        Parser.checkHasArgs(line, "Usage: %s <index>".formatted((isMarked) ? "mark" : "unmark"));
+
+        int index;
```
hzk-lab: It seems that you introduce many try-catch bolcks to read the next line. Why not consider packing this logic in a separate method to avoid repetition?

PR #38, opened 2025-01-23
[veehz] iP

veehz:
# Maid-Chan

> "Welcome Home" - Maid Chan (maybe) ([source](https://www.maidchan.com))

Modelled after a character in an [anime](https://sakurasou.fandom.com/wiki/Maid), Maid-Chan is

- Text-based
- ~Super~ _Smart_ and **Cool**

All you need to do is,

1. Download it from [here](https://github.com/veehz/ip/releases)
2. double-click it
3. add your tasks
4. chat with it :100: 

Future work:

- [X] Delete tasks
- [ ] Reminders (coming 

---

The `MaidChan` class is written like:

```java
public class MaidChan {
    private static TaskList taskList;
    private static Ui ui;
    private static Storage storage;
    private static Parser parser;
    // methods omitted
}
```


src/main/java/maidchan/task/TaskDateTime.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Parses a string into a LocalDateTime object.
+ */
+public class TaskDateTime {
+    public static final String ACCEPTED_FORMATS = "yyyy-MM-dd HH:mm or yyyy-MM-dd";
+    private boolean hasTime;
```
Zonnie-23, line 13: Nice use of boolean sounding variable

Zonnie-23 (commented):
Overall, code was well organised and easily understandable. OOP was well planned and executed. JavaDoc was also comprehensive. Error messages are also insightful

src/main/java/maidchan/MaidChan.java
```diff
@@ -0,0 +1,58 @@
...
+public class MaidChan {
+    private final TaskList taskList;
+    private final Storage storage;
+    private final Parser parser;
+
+    /**
+     * Constructs a new MaidChan bot.
+     */
+    public MaidChan() {
+        storage = new Storage(System.getProperty("user.dir") + File.separator + "data"
```
Zonnie-23, line 24: I am not sure, but should the System class be imported?

src/main/java/maidchan/storage/Storage.java
```diff
@@ -0,0 +1,99 @@
...
+        return tasks;
+    }
+
+    /**
+     * Saves the todo list to the file.
+     * If the file does not exist, it will be created.
+     *
+     * @param tasks the list of tasks to save
+     */
+    public void saveTodoList(List<Task> tasks) {
```
Zonnie-23, line 91: nice use of verbs to name functions

Zonnie-23 (commented):
Just to add a few more comments

src/main/java/maidchan/MaidChan.java
veehz: Oh, this is interesting. I've never wondered that myself.

It doesn't need to be imported because according to the [Java Specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-7.html),

> A compilation unit automatically has access to all types declared in its package and also automatically imports all of the public types declared in the predefined package `java.lang`.

And [`System`](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html) is defined under `java.lang`.

PR #40, opened 2025-01-23
[Somneel Kumar Saha] iP

SomneelSaha2042:
# 🚀 TaskMaster: Enhancements & Updates

### 📝 Task Management Enhancements
- ✅ Implemented the `delete` command to remove tasks by index.
- ✅ Added the `help` command to display available commands and their usage.
- 🛠️ Improved error handling for:
  - Invalid commands
  - Out-of-range task indices

### 🧪 Test Cases
- 📄 Created `input.txt` and `EXPECTED.TXT` with test cases for:
  - `todo`
  - `deadline`
  - `event`
  - `list`
- 📊 Verified output correctness by comparing `ACTUAL.TXT` with `EXPECTED.TXT`.

### 📖 README Update
- 🛠️ **Installation & Running Instructions**
- 📜 **List of Available Commands & Descriptions**
- 🗂️ Example usage for `TaskMaster`

---

## **Setup & Usage Instructions**
1. Clone the repository:
   ```sh
   git clone https://github.com/yourusername/taskmaster.git
   cd taskmaster


src/main/java/taskmaster/parser/Parser.java
```diff
@@ -0,0 +1,163 @@
...
+                return LocalDateTime.parse(dateTimeString, formatter);
+            } catch (DateTimeParseException ignored) {
+                // Try the next format
+            }
+        }
+        throw new TaskMasterException(
+                "Invalid date format: " + dateTimeString
+                        + ". Supported formats are: d/M/yyyy HHmm, d-M-yyyy HHmm, ISO_LOCAL_DATE_TIME."
+        );
+    }
```
benedictleejr, line 113: # Ignoring Exceptions, is this the best way to handle it?
I think having an empty catch block could potentially lead to bugs that are hard to catch down the road 🐛 

An alternative approach could be to first check and verify the format of the input string, and then parse using the appropriate format

src/main/java/taskmaster/commands/ExitCommand.java
```diff
@@ -0,0 +1,32 @@
...
+        return "Goodbye! Hope to see you again soon.";
+    }
+
+    /**
+     * Indicates that this command terminates the application.
+     *
+     * @return {@code true} since this command exits the application.
+     */
+    @Override
+    public boolean isExit() {
```
benedictleejr, line 29: # Is there a reason this is extracted out into a function?
I feel that there isn't a need to write functions for all the command classes to return a true/false based on whether to exit the program, especially since only the ExitCommand class is supposed to exit.
I'm curious to know if there's a reason for this design 😄 

Alternatively, you could store this as a a boolean attribute, or even handle the logic for exiting in your main app runner class.

benedictleejr (commented):
LGTM 😉  just some questions and suggestions on code design!

src/main/java/taskmaster/parser/Parser.java
SomneelSaha2042: Hey thanks for catching that! I shall resolve it ASAP

src/main/java/taskmaster/commands/ExitCommand.java
SomneelSaha2042: Its so that isExit can be called on an object of type Command from another class.

PR #41, opened 2025-01-23
[itsvari] iP

itsvari:
# SONGBIRD™ AI Assistant Mark IV 🤖 
### by VariTech Heavy Industries, ©️ 3025. All rights reserved and enforced by nanite swarm.

> "The only AI assistant with full military-grade neural encryption and integrated thought synchronization capabilities." - Dr. Elena Volkov, Chief of Cybernetic & Autonomic Research

Welcome to the official repository of SONGBIRD™, your personal AI productivity enhancement system. This state-of-the-art digital consciousness has been carefully engineered by VariTech's elite team of neural architects to serve your every ~~surveillance~~ organizational need.

## Features
* **Military-grade** task management protocols
* _Quantum-encrypted_ storage capabilities
* Advanced ~~mind control~~ reminder systems
* Multi-dimensional datetime parsing
* Neural-adaptive user interface

## System Requirements
1. Minimum quantum processing unit: Mark VII or higher
1. Neural interface port (~~mandatory~~ optional)
1. Minimum 1TB neural bandwidth
1. VariTech-approved consciousness stabilizer

## Quick Start
```java
    Songbird songbird = new Songbird();
    songbird.init();  // Initializes neural link
}
```

## Implementation Status
* [x] Basic consciousness implementation
* [x] Task management protocols
* [x] Storage encryption
* [ ] Full neural integration
* [ ] Quantum teleportation capabilities
* [ ] Time manipulation features

## Safety Protocols
* Never look directly into SONGBIRD's digital consciousness. Any damage to one's sanity, loss of will, or death will not be covered under the Limited Warranty.
* Maintain regular neural backups
* Keep your [VariTech Mental Firewall™](https://itsvari.com) active at all times
* Use command `bye` to safely disconnect neural interface

## Legal Disclaimer
By using SONGBIRD™, you agree to allow VariTech Heavy Industries access to your:
* **Thoughts**
* _Dreams_
* ~~Hopes~~
* Memories
* ~~Soul~~

For technical support, please direct your thoughts to our neural support network or consult your assigned VariTech consciousness technician.

---
_VariTech Heavy Industries - Resistance Is Obsolete_


src/main/java/songbird/parser/DateTimeParser.java
```diff
@@ -0,0 +1,188 @@
...
+import songbird.exception.SongbirdMalformedCommandException;
+
+/**
+ * Helper class to parse date/time strings into LocalDate or LocalDateTime objects.
+ * Allows multiple input formats. Outputs are always in ISO date/time.
+ *
+ * @author Ashe Low
+ * @version CS2103T AY24/25 Semester 2
+ */
+public class DateTimeParser {
```
samuelneo, line 22: I like your implementation of the date/time parser, it's very robust and lenient towards various input formats

src/main/java/songbird/task/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+        return this.tasks.get(index);
+    }
+
+    /**
+     * Returns all tasks in the task list.
+     *
+     * @return A copy of the task list.
+     */
+    List<Task> getAll() {
```
samuelneo, line 109: is this method intentionally package-private? I think (unless Ctrl-F is failing me) it's only accessed once and within this class, so maybe it could be private?

src/main/java/songbird/parser/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+                    eventDescription = fromParts[0].trim();
+                    toParts = fromParts[1].split(" /to ", 2);
+                    eventStartString = toParts[0].trim();
+                } else {
+                    // if /to comes before /from
+                    eventStartString = fromParts[1].trim();
+                    toParts = fromParts[0].split(" /to ", 2);
+                    eventDescription = toParts[0].trim();
+                }
+                eventEndString = toParts[1].trim();
```
samuelneo: personally I found this section easier to parse using regex (though what you've done is definitely fine too, and it's nice that you've added support for having `/to` before `/from`)

src/test/java/songbird/storage/StorageTest.java
```diff
@@ -0,0 +1,293 @@
...
+     */
+    @AfterEach
+    public void tearDown() {
+        File file = testFilePath.toFile();
+        if (file.exists()) {
+            try {
+                // Ensure file handle is released by forcing GC and waiting a bit
+                System.gc();
+                Thread.sleep(100);
+                boolean deleted = file.delete();
```
samuelneo: coding standard: should be named to sound like a boolean (something like `isDeleted`)
this appears in a couple other places too (lines 81 and 246 of `StorageTest.java`)

samuelneo (commented):
very well-written and well-documented code with thorough tests; I was struggling to come up with comments

src/test/java/songbird/storage/StorageTest.java
itsvari: good catch lol, i wrote this code at ~6am at the end of a 7-hour sprint while extremely sleep-deprived haha ^-^

src/main/java/songbird/parser/DateTimeParser.java
itsvari: haha thanks! it definitely doesn't cover every format i want it to, i'll probably switch to using something like [hawking](https://github.com/zoho/hawking) instead. after all, someone else has done all the legwork, might as well use that ;)

src/main/java/songbird/parser/Parser.java
itsvari: regex would work too! but the regex required to handle fuzzy from/to matching is.... yuck:
`(.+?)\s+/(?:(?:from\s+(.+?)\s+/to)|(?:to\s+(.+?)\s+/from))\s+(.+)`

overall i felt that doing string splitting was a bit easier to comprehend, and also provides more extensibility if we want to detect alternate keywords in the future!

src/main/java/songbird/task/TaskList.java
itsvari: it's used in `TaskListTest::testGetAllEncapsulation`, so that's why it's package-private!

```java
/**
 * Tests the encapsulation of the getAll method.
 * Verifies that modifications to the returned list don't affect the original list.
 */
@Test
public void testGetAllEncapsulation() {
    // Arrange
    Task task = new ToDoTask("Read a book");
    taskList.addTask(task);


    // Act
    List<Task> allTasks = taskList.getAll();


    // Attempt to modify the returned list
    allTasks.add(new ToDoTask("New Task"));


    // Assert
    assertEquals(1, taskList.getSize(),
            "Modifying the returned list should not affect the original TaskList.");
}
```

samuelneo: ahh I see

src/main/java/songbird/parser/Parser.java
samuelneo: yeahh the regex gets messier if you're supporting fuzzy from/to matching haha

src/main/java/songbird/command/Command.java
```diff
@@ -0,0 +1,50 @@
...
+    public String getDescription() {
+        return description;
+    }
+
+    // ABSTRACT METHODS
+
+    /**
+     * Executes the command.
+     */
+    public abstract void execute();
```
zihung20, line 49: I like how you designed your Command class. Each subclass of Command implements `execute()`, ensuring they have their own behavior.

src/main/java/songbird/parser/DateTimeParser.java
```diff
@@ -0,0 +1,188 @@
...
+     * @return The LocalDateTime object parsed from the input string.
+     * @throws SongbirdMalformedCommandException If the input string cannot be parsed into a LocalDateTime object.
+     */
+    public static LocalDateTime parseDateTime(String input) throws SongbirdMalformedCommandException {
+        // try full date-time patterns
+        for (String pattern : DATE_TIME_PATTERNS) {
+            try {
+                return LocalDateTime.parse(input, DateTimeFormatter.ofPattern(pattern));
+            } catch (DateTimeParseException e) {
+                // ignore and try next pattern
```
wzhua02: Perhaps include a continue here? In case you add something after try catch? But not wrong here.

src/main/java/songbird/task/Task.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Represents a Task that the user wants to save.
+ * <p>
+ * Tasks can either be done or not done, and must have a description.
+ *
+ * @author Ashe Low
+ * @version CS2103T AY24/25 Semester 2
+ */
+public abstract class Task implements Serializable {
```
wzhua02, line 14: Was wondering why do you implement Serializable and what's its purpose?

src/main/java/songbird/command/CommandType.java
```diff
@@ -0,0 +1,46 @@
...
+
+import songbird.exception.SongbirdInvalidCommandException;
+
+/**
+ * Represents the different types of commands that the user can input.
+ *
+ * @author Ashe Low
+ * @version CS2103T AY24/25 Semester 2
+ */
+public enum CommandType {
```
wzhua02, line 11: Wondering why do you specify a command type class when each command already has a class?

src/main/java/songbird/parser/Parser.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * Parses the user input into a command object.
+     * The command object is created based on the user input.
+     *
+     * @param input The user input to be parsed.
+     * @return The command object that corresponds to the user input.
+     */
+    public Command parse(String input) throws SongbirdException {
```
wzhua02, line 52: I like the switch cases done in the parser.

src/main/java/songbird/Songbird.java
```diff
@@ -0,0 +1,102 @@
...
+    private Parser parser;
+    private TaskList tasks;
+
+    /**
+     * Constructs the Songbird chatbot.
+     */
+    public Songbird() {
+        try {
+            this.ui = new Ui();
+            Storage storage = new Storage("./data/tasklistDB");
```
wzhua02, line 32: Perhaps can move the filepath into the storage class?

src/test/java/songbird/task/DeadlineTaskTest.java
```diff
@@ -0,0 +1,50 @@
...
+    private DeadlineTask deadlineTask;
+    private final LocalDateTime deadline = LocalDateTime.of(2025, 10, 30, 23, 59);
+
+    @BeforeEach
+    public void setUp() {
+        deadlineTask = new DeadlineTask("Submit report", deadline);
+    }
+
+    @Test
+    public void testConstructor() {
```
wzhua02: Perhaps change the test method name format to "unitBeingTested_descriptionOfTestInputs_expectedOutcome". Also could include some javadocs.

wzhua02 (commented):
The code is really well written, good job.

src/main/java/songbird/Songbird.java
itsvari: It's for extensibility, to allow multiple Storage instances to be initialised. Better not to hardcode magic/hidden filepaths into the Storage class itself

src/main/java/songbird/task/Task.java
itsvari: You need to implement Serializable to use Java's built in SerDes operations, in order to write it to a file

src/main/java/songbird/command/CommandType.java
itsvari: I find it much cleaner to use an enum combined with a switch, rather than constantly doing `InstanceOf` checks and chained if-else statements, when doing command processing.

IIRC, Java 17 does have support for type pattern matching, but only as a preview. I just went with an enum instead.

src/test/java/songbird/task/DeadlineTaskTest.java
itsvari: yup, good catch

src/main/java/songbird/parser/DateTimeParser.java
itsvari: Continue is not needed here, even if we were to place something after the try/catch block; execution would just flow on further down.

src/main/java/songbird/parser/DateTimeParser.java
```diff
@@ -0,0 +1,188 @@
...
+        "dd-MM-yyyy h:mma",
+        "dd/MM/yyyy h:mma",
+        "dd-MM-yyyy h:mm a",
+        "dd/MM/yyyy h:mm a",
+        "dd-MM-yyyy hh a",
+        "dd/MM/yyyy hh a",
+        "dd-MM-yyyy h a",
+        "dd/MM/yyyy h a",
+        "MM/dd/yyyy HH:mm",
+        "MM-dd-yyyy HH:mm",
```
zihung20: Wouldn't there be cases where months are misinterpreted? For example, if a user enters 11-01-2025 12:34 intending it as November 1st, but due to the sequence, it gets interpreted as January 11th instead? Or this is your intentional implementation?

zihung20 (commented):
Overall, I found this PR flawless. It’s hard to find any issues since exceptions are well handled, the files are well-structured, and the test cases comprehensively cover the class behavior. **Very good work!** 😊

src/main/java/songbird/parser/DateTimeParser.java
itsvari: yep, that's a very good catch! it's hard to avoid ambiguity without providing a configuration option.

though, it could be customized so that it prioritizes the system locale's way of writing dates... 🤔

but yeah, i18n and l10n is *hard*.

PR #46, opened 2025-01-24
[ToxOptimism] iP

ToxOptimism:
# Aurora 
>"When does the comet become a meteor?
> When does a candle become a blaze?
> When does a man become a monster?"
> \- "Just A Man" EPIC: The Musical

## Aurora is a task management ~~application~~ ***chatbot*** that is
- text-based
- user-friendly
- easy to set-up

## Set-up
1. download the [Aurora](https://github.com/ToxOptimism/ip/releases/tag/A-Jar) release file.
2. double-click it or cd to the directory Aurora is downloaded and run java -jar aurora-v0.1.jar in the directory
3. add your tasks.
4. let Aurora handle the rest! :purple_heart:

## Aurora's suite of features:
- [x] Add todos, events, deadlines
- [x] List tasks
- [x] Mark tasks as done
- [x] Delete tasks
- [ ] Find tasks by description
- [ ] Use a graphical user interace

## Here's the main method of Aurora:
```java

    try {
        storage.generateTaskListFile();
        loadTaskList();
    } catch (AuroraException e) {
        Ui.printMsg(e.getMessage());
        return;
    }

    Scanner sc = new Scanner(System.in);
    Command command = null;

    // Greet User
    Ui.printMsg(GREETING);

    // Chatbot
    while (command == null || !command.isExitCommand()) {
        String input = sc.nextLine().trim();
        try {
            command = Parser.of().parseCommand(input);
            command.execute(taskList, storage);
        } catch (AuroraException e) {
            Ui.printMsg(e.getMessage());
        }
    }

}
```

*You can take a look at Aurora's github [here](https://github.com/ToxOptimism/ip).*


src/main/java/aurora/command/AddCommand.java
```diff
@@ -0,0 +1,29 @@
...
+public abstract class AddCommand extends Command {
+
+    /**
+     * Adds a task to the task list and prints a success message.
+     *
+     * @param t the task to be added.
+     * @param taskList the taskList that stores the task.
+     * @param storage the storage to write to.
+     * @throws AuroraException if appending to storage fails.
+     */
```
iamanoob44: I like how informative the JavaDoc comments are, nice!

src/main/java/aurora/command/AddCommand.java
```diff
@@ -0,0 +1,29 @@
...
+    public void addToList(Task t, TaskList taskList, Storage storage) throws AuroraException {
```
iamanoob44: Perhaps it would be good if you can rename your variable name 't' to something more descriptive such as 'taskToAdd' so that it follows the coding standard more closely? Overall, I still find your code easy to read and understand!

Since we usually allocate one single letter : `i`,`j` or `k` for for loops

src/main/java/aurora/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,97 @@
...
+        // If there is no /by
+        if (byDateStart == -1) {
+            throw new AuroraException("Missing argument: \"/by By\"."
+                    + "\nUsage: \"deadline Description /by By\"");
+
+        // If there is no details after /to
+        } else if (byDateStart + 3 == info.length()) {
+            throw new AuroraException("Missing argument: \"By\" in \"/by By\".\n"
+                    + "Usage: \"deadline Description /by By\"");
+        }
```
iamanoob44: Perhaps you can define certain repetitive hard-coded strings as constants to help increase your code maintainability and align with the coding standards more closely? For example, perhaps you can consider:

`private static final String BY_KEYWORD = "/by";`
`private static final String DEADLINE_DESCRIPTION = "Usage: \"deadline Description /by By\"";`

src/main/java/aurora/command/AddEventCommand.java
```diff
@@ -0,0 +1,135 @@
...
+        } else if (fromDateStart + 5 == beforeTo.length()) {
+            throw new AuroraException("Missing argument: \"From\" in \"/from From\".\n"
+                    + "Usage: \"event Description /from From /to To\"");
+
+        // If there is no details after /to
+        } else if (toDateStart + 3 == info.length()) {
+            throw new AuroraException("Missing argument: \"To\" in \"/to To\".\n"
+                    + "Usage: \"event Description /from From /to To\"");
+        }
+
```
iamanoob44: I noticed a similar issue found in the `AddDeadlineCommand.java` file earlier. Perhaps you can try to define constants too for certain repetetive hard-coded strings?

src/main/java/aurora/exception/AuroraException.java
```diff
@@ -0,0 +1,16 @@
...
+
+    /**
+     * Constructs a new AuroraException with the specified message.
+     *
+     * @param message the error message detailing the cause of the exception.
+     */
+    public AuroraException(String message) {
+        super(message);
+    }
+}
```
iamanoob44, line 16: LGTM, I like how you included javadoc comments for both the class and also for the constructor. I also like how clean your code is for this file!

src/main/java/aurora/task/TaskList.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param index the 1-based index of the task to get.
+     * @return Task the task at specified index.
+     * @throws AuroraException if the index is out of bounds.
+     */
+    public Task getTask(int index) throws AuroraException {
+
+        validateIndex(index); // throws an exception
+
+        return taskList.get(index - 1);
+    }
```
iamanoob44, line 44: Perhaps you can incorporate a 'try-catch' block to capture the `AuroraException`. Any reason why you did it this way instead of using a 'try-catch' ?

iamanoob44 (commented):
LGTM, just some nits to fix. Overall, great work with the code! it still closely aligns with the coding standards as I do not see any blatant violations of the coding standards for the course.

src/main/java/aurora/task/TaskList.java
ToxOptimism: Ah, the reasoning is that it is not the responsibility of the `getTask` function to handle the `AuroraException` when it is the caller that is supplying an invalid index. 

As such this exception is actually thrown up to any command that uses this function and in turn the command will typically throw it to the chatbot, since its also not the command's responsibility to handle invalid input. Thus the chatbot is the one to handle the try-catch capturing of the AuroraException.

Hope that clarifies things, and do you have any recommendation for comment improvements to bring additional clarity to this issue within the code?

src/main/java/aurora/command/AddCommand.java
ToxOptimism: I will have to look into it, as due to the small scope of the function, a short name is appropriate for this situation. Additionally, this is standard with other methods such as TaskList's markTaskAsDone which references `Task t` as a scratch variable.  

Yes `i`, `j`, `k` are scratch variables usually reserved for for loops, likewise `m`, `n` are usually reserved for `int` input parameters and `c` and `d` references characters. As such, the code will likely be using `t` as a scratch variable for any variable of datatype Task.

src/main/java/aurora/command/AddDeadlineCommand.java
ToxOptimism: Thanks for the suggestion! 😄 

Will likely implement this in future commits.

src/main/java/aurora/command/AddCommand.java
iamanoob44: Ahh i see i see with regards to the scratch variable, but i wont deny the fact that the code still looks clean though. It's just i think perhaps is better to rename the variable 't' to 'task' for better clarity, so that it better aligns with the coding standard i guess?

src/main/java/aurora/task/TaskList.java
iamanoob44: Ohhh i see, thanks for the clarification though and my bad on this then!

and with regards to the comment improvements, perhaps you can include your "validateIndex" part into the JavaDoc comments.

So perhaps something like this?

```Java
/**
 * Validates if the 1-based index is within the bounds of the list.
 * <p>
 *     The validateIndex method in {@code TaskList} also throws AuroraException when ...
 * </p>
 * @param index the 1-based index to be validated.
 * @throws AuroraException if the index is out of bounds.
 */
 ```

File: src/main/java/aurora/command/AddDeadlineCommand.java [hunk 7]
```diff
@@ -0,0 +1,97 @@
...
+import aurora.task.TaskList;
+import aurora.util.Parser;
+
+/**
+ * Represents a command to add a Deadline to the TaskList.
+ */
+public class AddDeadlineCommand extends AddCommand {
+
+    // Deadline specific fields
+    private LocalDateTime bDate;
```
jeronkk: Maybe instead of bDate try byDate

src/main/java/aurora/command/AddEventCommand.java
```diff
@@ -0,0 +1,135 @@
...
+import aurora.task.TaskList;
+import aurora.util.Parser;
+
+/**
+ * Represents a command to add an Event to the TaskList.
+ */
+public class AddEventCommand extends AddCommand {
+
+    // Event specific fields
+    private LocalDateTime fDate;
```
jeronkk: Consider renaming it to fromDate and toDate

src/main/java/aurora/command/AddToDoCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param taskList the taskList to add to.
+     * @param storage the storage to write to.
+     * @throws AuroraException if an error occurs in lower-level method.
+     */
+    @Override
+    public void execute(TaskList taskList, Storage storage) throws AuroraException {
+
+        super.execute(taskList, storage);
+
+        ToDo td = new ToDo(description);
```
jeronkk: maybe instead of td consider toDoTask

src/main/java/aurora/command/DeleteCommand.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param taskList the taskList to delete from.
+     * @param storage the storage to overwrite with taskList data.
+     * @throws AuroraException if an error occurs in lower-level method.
+     */
+    @Override
+    public void execute(TaskList taskList, Storage storage) throws AuroraException {
+
+        super.execute(taskList, storage);
+
+        Task t = taskList.deleteFromList(index); // throws AuroraException if index is out of bounds
```
jeronkk: Can be named task instead of t

src/test/java/aurora/io/UiStub.java
```diff
@@ -0,0 +1,41 @@
...
+        isSingletonSet = true;
+    }
+
+    public void setAurora(Aurora aurora) {
+        this.aurora = aurora;
+    }
+
+    public void printMsg(String msg) {
+        return;
+    }
```
jeronkk, line 35: Maybe write out the full word printMessage()

jeronkk (commented):
In all really impressed with your chatbot. Other than small naming conventions, really solid work!

src/main/java/aurora/Aurora.java
```diff
@@ -0,0 +1,91 @@
+package aurora;
```
bhavinaa, line 1: This is my first PR, regarding the coding standard alone. So I would only be commenting on that. 

I like how you have organized your class, while grouping related packages, and at the same time you have restricted the accessibility, such that it is only accessible

src/main/java/aurora/command/AddDeadlineCommand.java
bhavinaa: Perhaps you could consider renaming the variable bDate. It is not very intuitive as to what that refers to. It would be hard for others to understand the code, the second time you see it

src/main/java/aurora/command/AddEventCommand.java
```diff
@@ -0,0 +1,135 @@
...
+    private LocalDateTime tDate;
```
bhavinaa: I really like the javDoc comments, they are very descriptive and then explains the code well

src/main/java/aurora/command/AddToDoCommand.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Parses the arguments for the command.
+     *
+     * @param argsList the arguments to parse.
+     * @throws AuroraException the appropriate exception message if unable to parse arguments.
+     */
+    @Override
+    public void parseArgs(String[] argsList) throws AuroraException {
```
bhavinaa, line 51: perhaps you could consider abstracting out the parsing to a different class, to reduce code repetition

src/main/java/aurora/util/Parser.java
```diff
@@ -0,0 +1,167 @@
+package aurora.util;
```
bhavinaa, line 1: I like how you have specifically imported the functions you use from the common libraries instead of doing it with the *

bhavinaa (commented):
Overall, I feel you have really good code quality. You have consistent naming, and then your JavaDocs is really descriptive  making it easy for the new readers to look at it

PR #48, opened 2025-01-24
[TeaMakesMePee] iP

TeaMakesMePee:


src/main/java/tiffy/Tiffy.java
```diff
@@ -0,0 +1,106 @@
...
+            } catch (TiffyException te) {
+                UiManager.getInstance().printException(te);
+            }
+        } catch (IndexOutOfBoundsException e) {
+            throw new TiffyException("Invalid index!",
+                    TiffyException.ExceptionType.INVALID_INDEX, e);
+        }
+    }
+
+    public static void handleRequests(Parser p, String input, TaskManager tm) throws TiffyException {
```
nicholasohjj: `handleRequests` could be more descriptive – It is responsible for parsing input and executing commands, so something like `processUserCommand `or `executeCommand `might be clearer.

src/main/java/tiffy/Tiffy.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Generates a response for the user's chat message.
+     */
+    public String getResponse(String input) {
+        return "Duke heard: " + input;
+    }
+
+    public static void markDoneUndone(List<Task> tasks, boolean mark, int index) throws TiffyException {
+        try {
+            Task temp = tasks.get(index - 1);
```
nicholasohjj: "temp" is too generic. A more descriptive name like `taskToUpdate `or `matchedTasks` would improve readability.

src/main/java/tiffy/Tiffy.java
```diff
@@ -0,0 +1,106 @@
...
+                UiManager.getInstance().printException(te);
+            }
+        } catch (IndexOutOfBoundsException e) {
+            throw new TiffyException("Invalid index!",
+                    TiffyException.ExceptionType.INVALID_INDEX, e);
+        }
+    }
+
+    public static void handleRequests(Parser p, String input, TaskManager tm) throws TiffyException {
+        String[] partition = p.handleRequests(input);
```
nicholasohjj: a clearer name like `commandParts `or `parsedInput `would better reflect its purpose.

src/main/java/tiffy/Tiffy.java
nicholasohjj: Single-character variable names should be avoided unless widely understood (i, n). A more descriptive name like `parser `would be clearer.

nicholasohjj (changes requested):
Good job and keep up the work. some nits to fix

src/main/java/tiffy/Tiffy.java
```diff
@@ -0,0 +1,106 @@
...
+            case "list" -> {
+                if (tm.getTasks().isEmpty()) {
+                    throw new TiffyException("You have no tasks! Add some.",
+                            TiffyException.ExceptionType.ZERO_TASK);
+                }
+                UiManager.getInstance().printTasks(tm.getTasks());
+            }
+            case "todo" -> {
+                tm.addTask(new Todo(partition[1]));
+            }
```
GZTan23: I believe that the `case` statements should not be indented?

src/main/java/utility/Parser.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Processes the user's request and splits it into components for further handling.
+     *
+     * @param request The user's input string.
+     * @return An array of strings representing the parsed command and its arguments.
+     * @throws TiffyException If the input is invalid or incomplete.
+     */
+    public String[] handleRequests(String request) throws TiffyException {
+        String[] partition = request.split(" ");
```
GZTan23: I think that 'partition' should be 'partitions' instead since it signifies a collection of Strings

src/main/java/task/Task.java
```diff
@@ -0,0 +1,97 @@
+package task;
+
+import exception.TiffyException;
+import manager.UiManager;
+
+/**
+ * Represents an abstract Task with a description and a completion status.
+ * This class serves as a base for specific types of tasks.
+ */
+public abstract class Task {
```
GZTan23, line 10: Awesome job in making this an abstract class rather than a normal class!

src/main/java/task/Todo.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Returns the formatted string representation of the todo task.
+     *
+     * @return Formatted string representing the todo task.
+     */
+    @Override
+    public String getFormattedTask() {
+        return "T|" + this.isDone + "|" + this.description;
+    }
```
GZTan23, line 37: Maybe the JavaDoc is not required here? Since you've already included it in the Task class, which this method overrides and the functionality of the method is the same as listed in the Task class.

GZTan23 (commented):
Great job on the code, seems like you've put in quite the effort! All the best for the rest of the iP!

PR #55, opened 2025-01-24
[zwliew] iP

zwliew:
# Peepo

Peepo is a task tracker. What more do you want? Move along if you're not interested.

## Features

- Track Events with a start and end date
- Track Todos
- Track Deadlines

## Installation steps

1. Download the JAR file [here](https://github.com/zwliew/ip/releases/tag/A-Jar)
2. Run the app with the command `java -jar peepo.jar`.


src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public final class Task {
+  protected boolean isDone;
```
githubcsuser: Since Task is final, it cannot be subclassed, should these fields be private instead of protected to enforce encapsulation?

src/main/java/Peepo.java
```diff
@@ -11,17 +11,35 @@ public static void main(String[] args) {
...
+    final ArrayList<Task> texts = new ArrayList<>();
 
     try (Scanner scanner = new Scanner(System.in)) {
       for (String input = scanner.nextLine(); !"bye".equals(input); input = scanner.nextLine()) {
         System.out.println(LINE);
         if ("list".equals(input)) {
           for (int i = 0; i < texts.size(); i++) {
-            System.out.println(INDENT + (i + 1) + ". " + texts.get(i));
+            final var task = texts.get(i);
+            System.out.println(INDENT + (i + 1) + ". [" + task.getStatusIcon() + "] " + task.description);
```
githubcsuser: Would it be better to use a getter method (task.getDescription()) instead of directly accessing description to follow encapsulation principles?

src/main/java/Peepo.java
```diff
@@ -11,17 +11,35 @@ public static void main(String[] args) {
...
           }
+        } else if (input.startsWith("mark ")) {
+          final var idx = Integer.parseInt(input.substring(5)) - 1;
```
githubcsuser: Would adding a check for valid input length before calling substring(5) prevent potential StringIndexOutOfBoundsException?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public final class Deadline extends Task {
+  protected String by;
```
githubcsuser: Since Deadline and Event are final and cannot be subclassed, should these fields be private instead of protected to enforce encapsulation?

githubcsuser (commented):
Just some minor comments to consider.

PR #59, opened 2025-01-24
[samuelneo] iP

samuelneo:
## Carbon

> * 1. - [x] This [sentence](https://en.wikipedia.org/wiki/Sentence_(linguistics)) _exists_ 🌱 to fulfil `formatting` requirements

```java
// and so does this code block
```


src/main/java/carbon/task/Temporal.java
```diff
@@ -0,0 +1,216 @@
...
+ */
+public class Temporal implements Comparable<Temporal> {
+    enum TemporalType {
+        DATE,
+        TIME,
+        DATETIME,
+        TEXT
+    }
+
+    private static final DateTimeFormatter[] DATE_FORMATTERS = {
```
wzhua02: I like that you have a class dedicated to parse various formats of dates and time.

src/main/java/carbon/task/Todo.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public class Todo extends Task {
+    /**
+     * Creates a Todo with the specified description.
+     *
+     * @param description Description of the Todo.
+     */
+    public Todo(String description) {
+        super('T', description);
+    }
```
wzhua02, line 16: Maybe can have the methods like toString and getStorageText for Todo objects implemented here? 
My thoughts are that the ToDo class seems redundant to have.

wzhua02 (commented):
The code is well written, feel like I've learnt a lot just from reading your code.

src/main/java/carbon/task/Todo.java
samuelneo: maybe, but I don't really see a need to include the implementation here since `Todo` can just use the implementation in `Task`

src/main/java/carbon/gui/MainWindow.java
```diff
@@ -0,0 +1,72 @@
...
+     * to the dialog container. Clears the user input after processing.
+     */
+    @FXML
+    private void handleUserInput() {
+        String input = userInput.getText();
+        String response = carbon.getResponse(input);
+        String command = carbon.getCommand();
+        dialogContainer.getChildren().addAll(
+                DialogBox.getUserDialog(input, userImage),
+                DialogBox.getCarbonDialog(response, carbonImage, command)
```
loyaltypollution, line 66: Good code split that is clear and in line with given conventions!

src/main/java/carbon/task/Temporal.java
```diff
@@ -0,0 +1,216 @@
...
+ *     <li>Text (none of the above)</li>
+ * </ul>
+ */
+public class Temporal implements Comparable<Temporal> {
+    enum TemporalType {
+        DATE,
+        TIME,
+        DATETIME,
+        TEXT
+    }
```
loyaltypollution: Could I ask why the implementation of Temporal instead of the more encompassing java.time.LocalDateTime?

src/main/java/carbon/utils/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+     * For example, if the user inputs "deadline assignment /by 1pm",
+     * <code>text</code> would be "assignment /by 1pm".
+     *
+     * @param text Details of the Deadline.
+     * @return The Deadline object.
+     */
+    public static Deadline parseDeadline(String text) {
+        // Regex matches a String of the form "{A} /by {B}", where A and B each contain at least
+        // one non-whitespace character
+        Matcher matcher = Pattern.compile("^\\s*(\\S.*?)\\s+/by\\s+(\\S.*?)\\s*$").matcher(text);
```
loyaltypollution, line 29: Perhaps the regex string could be made private static final in Parser?
This gives us an opportunity to name the regex too.

src/main/java/carbon/utils/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+        if (!matcher.find()) {
+            throw new InvalidArgumentException(
+                    "Deadline commands should be formatted as \"deadline [description] /by [due date/time]\"");
+        }
+        return new Deadline(matcher.group(1), matcher.group(2));
+    }
```
loyaltypollution, line 35: However, using regex groups to parse the deadline directly is great!

src/main/java/carbon/utils/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+                            String from = scanner.nextLine().trim();
+                            String to = scanner.nextLine().trim();
+                            yield new Event(description, from, to);
+                        }
+                        default -> throw new InvalidFileFormatException();
+                    };
+                    if (isDone) {
+                        task.markAsDone();
+                    }
+                    taskList.add(task);
```
loyaltypollution: Good job with the parsing of Storage file!

src/main/java/carbon/utils/Ui.java
```diff
@@ -0,0 +1,77 @@
...
+    private final TaskList taskList;
+    private String mostRecentCommand;
+    private final String startMessage;
+
+    /**
+     * Creates a new UI process.
+     */
+    public Ui() {
+        this.taskList = new TaskList();
+        startMessage = Storage.loadDataFile(taskList) + "Hello! What can I do for you?";
```
loyaltypollution: For me, it is not very clear what startMessage refers to. I get confused between start of JVM initialization, start of GUI construction or start of chatbot.

Is welcomeMessage clearer?

src/main/java/carbon/utils/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+     * @return Message describing any abnormalities (empty if none).
+     */
+    public static String loadDataFile(TaskList taskList) {
+        String message = "";
+
+        File dataFile = new File(TASKS_FILE_PATH);
+        if (!dataFile.getParentFile().exists() && !dataFile.getParentFile().mkdirs()) {
+            throw new RuntimeException("Unable to create data/user/ directory");
+        }
+        boolean fileExists;
```
loyaltypollution: As per [convention given](https://se-education.org/guides/conventions/java/intermediate.html#variables), we could use verb+noun. 
```suggestion
        boolean hasFile;
```
Is "hasFile" clearer?

loyaltypollution (commented):
LGTM. Just a few Nits to pick. Great job and amazing progress! 👍

src/main/java/carbon/task/Temporal.java
samuelneo: good question, I wanted an object that could hold either a date, time, datetime, or text, so that it's more lenient for the user (e.g. the user can just specify a date instead of a datetime if the time isn't important)

src/main/java/carbon/utils/Parser.java
samuelneo: yeah possibly, though I only use this regex once so I don't really see that as necessary (though if I end up using it again or extending my use of regex I'll probably do that)

src/main/java/carbon/utils/Ui.java
samuelneo: good point, I'll rename it to `welcomeMessage`

src/main/java/carbon/utils/Storage.java
samuelneo: nice catch, thanks! will correct it

PR #60, opened 2025-01-24
[leewaikin19] iP

leewaikin19:
# Glados
> “Remember Before When I Was Talking About Smelly Garbage Standing Around Being Useless? That Was A Metaphor. I Was Actually Talking About You.” - GlaDOS

Glados allows you to
- track your tasks
- add deadlines
- save your tasks locally!!!

To install and use Glados, follow the steps below:
1. Download the latest release from [here](https://github.com/leewaikin19/ip/releases)
2. Double click to open the application
3. Add and/or remove tasks

## Why use Glados? 
Because it is only ~~$99.99~~ **free**, *open-source* and **lightweight**. 

Features:
- [x] Add and manage tasks
- [ ] Online syncing (TBA)
- [ ] Reminders (TBA)

Here is a snippet of the `main` method:
```java
    new Glados("data/tasks.txt").run();
}
```


src/main/java/glados/ui/Parser.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Parses a string into a DateTime format. Returns null if string is not a date
+     * or time.
+     * 
+     * @param str String to be parsed
+     * @return LocalDateTime Parsed datetime
+     */
+    private static LocalDateTime parseDateTime(String str) {
+        String[] formats = { "d/M/yyyy", "d/M/yyyy HHmm", "d/M/yyyy hh:mm a", "d/M/yyyy hh:mm:ss a", "yyyy/M/d",
```
awc1116, line 26: Would it be better for you to create an enum for your dateTimeFormat?

src/main/java/glados/tasks/Task.java
```diff
@@ -0,0 +1,36 @@
+package glados.tasks;
+
+/** Task that can be stored in the task list */
+
+public class Task {
```
awc1116, line 5: Would it be better to make it abstract since you don't have to initiate a Task object?

src/main/java/glados/commands/Command.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Executes task
+     * 
+     * @param tasks   task list of the program
+     * @param ui      ui of the program
+     * @param storage local storage of the program
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
```
awc1116: Would it be cleaner to make this an abstract method and for each individual command create a separate command class so that it would be easier for you to add, edit or remove specific commands when needed.

src/main/java/glados/commands/AddTaskCommand.java
```diff
@@ -0,0 +1,116 @@
...
+import glados.tasks.Deadline;
+import glados.tasks.Event;
+import glados.tasks.Task;
+import glados.tasks.TaskList;
+import glados.tasks.Todo;
+import glados.ui.Ui;
+import glados.exceptions.CommandException;
+
+/** Command to add a task to the list */
+public class AddTaskCommand extends Command {
```
awc1116, line 15: Since Parser already checked for todo/deadline/event type, would it be easier for you to create 3 separate command classes to add each type of task? 

What if I have one more type of task with more variables? Would it take a lot of effort to change your current code to add in this task type?

src/main/java/glados/tasks/Task.java
```diff
@@ -0,0 +1,36 @@
...
+    protected String description;
+    protected boolean isDone;
```
e1093051, line 7: I like how you name the boolean variable so that its data type is told.

src/main/java/glados/tasks/Task.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    /**
+     * Returns 'X' if task is done, ' ' otherwise
+     * 
+     * @return String
```
e1093051: Would it be clearer to state what the return string represents?

src/main/java/glados/tasks/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+     * 
+     * @param index index
+     * @return Task task at index
+     */
+    public Task get(int index) {
+        return taskList.get(index);
+    }
+
+    /**
+     * Removes the ith task
```
e1093051: You probably miss the period.

src/main/java/glados/tasks/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+
+    public int size() {
+        return taskList.size();
+    }
+
+    /**
+     * Gets the ith task
+     * 
+     * @param index index
```
e1093051: You probably want to have the i of the second index to be capital.

e1093051 (commented):
Found some minor mistakes, but overall is pretty good!

src/main/java/glados/ui/Glados.java
```diff
@@ -0,0 +1,46 @@
...
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.loadData());
+        } catch (GladosException e) {
+            Ui.show(e.getMessage());
+            tasks = new TaskList();
+        }
+    }
+
+    public void run() {
```
e1093051, line 29: Would it be clearer to roughly explain this public method?

PR #76, opened 2025-01-25
[Lee De En] iP

leedeen01:
MavisPro frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/leedeen01/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [ ] Managing tasks
- [ ] Reminders (coming soon)

#
```java
public class Main {
    }
}
```


src/main/java/mavis/Mavis.java
```diff
@@ -0,0 +1,41 @@
...
+    private Ui ui;
+
+    /**
+     * Constructs a Mavis object with the specified file path.
+     * Initializes the UI, storage, and task list.
+     *
+     * @param filePath The file path to load and save tasks.
+     */
+    public Mavis(final String filePath) {
+        ui = new Ui();
```
Shamanbenny, line 34: Would it be better to have `Ui` be a `static` class? (That way you won't have to instantiate `Ui` each time you require it)

src/main/java/mavis/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+                throw new MavisException("Find command requires a search term.");
+            }
+            return new FindCommand(toFind);
+        } else if (input.equals("bye")) {
+            return new ExitCommand();
+        } else if (input.equals("list")) {
+            return new ListCommand();
+        } else {
+            throw new MavisException("I'm sorry, but I don't know what that means.");
+        }
```
Shamanbenny: I feel like it might be easier to grasp the entire structure of this if it was set in `switch case` instead?

Shamanbenny: If I am also not mistaken, I believe `switch case` on String variable is done with Hash lookup, which results in an O(1) lookup rather than chaining `if else` statements which could result in O(n), where `n` is the number of conditions.
Read more about it [here](https://medium.com/@thedevelopernote/efficiency-of-if-vs-switch-in-java-a-quick-comparison-32f4ee0e20a3).

src/main/java/mavis/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+            ToDo todo = new ToDo(name);
+            return new AddCommand(todo);
+        } else if (input.startsWith("deadline")) {
+            String[] parts = input.substring(8).trim().split("/", 3);
+            if (parts.length < 2) {
+                throw new MavisException("A deadline task must be in this format: task /by due date.");
+            }
+            String desc = parts[0].trim();
+            String byPart = parts[1].trim();
+            if (!byPart.toLowerCase().startsWith("by")) {
```
Shamanbenny: I like how you use `.startsWith()` to deal with the `/` commands. Quite an elegant and ideal solution!

src/main/java/mavis/Storage.java
```diff
@@ -0,0 +1,109 @@
...
+ * The Storage class is responsible for loading and saving tasks to and from a file.
+ * It handles task persistence, reading tasks from a specified file path, and saving
+ * tasks back to the file when updated.
+ */
+public class Storage {
+
+    /**
+     * The file path to the Mavis data file for this task manager.
+     */
+    private String filePath;
```
Shamanbenny, line 24: Should this be `final` to prevent accidental assignment? If so, reminder to use `SCREAMING_SAKE_CASE` as referenced [here](https://se-education.org/guides/conventions/java/intermediate.html#naming).

src/main/java/mavis/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+import mavis.task.Task;
+
+/**
+ * The TaskList class provides functionality to manage a collection of tasks.
+ * It allows performing various operations on tasks, such as:
+ *
+ * - Adding new tasks
+ * - Deleting tasks by their identifier
+ * - Marking tasks as completed or incomplete
+ * - Listing all tasks with their current status
```
Shamanbenny, line 14: Would it be ideal to remove "lists" from the class header/JavaDoc? I believe it doesn't display as intended when hovering the class containing said JavaDoc on IDEs.

src/main/java/mavis/Storage.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * The Storage class is responsible for loading and saving tasks to and from a file.
+ * It handles task persistence, reading tasks from a specified file path, and saving
+ * tasks back to the file when updated.
+ */
+public class Storage {
+
+    /**
+     * The file path to the Mavis data file for this task manager.
+     */
```
Shamanbenny, line 23: Refer to the [Java coding standard on 'Comments'](https://se-education.org/guides/conventions/java/intermediate.html#comments), would it be better to use single line comments on class members?

Shamanbenny (commented):
Wonderful work thus far, aside from a few nits. Keep up the good coding standard!

PR #79, opened 2025-01-25
[TiewJiaLiang] iP

TiewJiaLiang:
### WizT

> "The function of good software is to make the complex appear to be simple" - Grady Booch ([source](https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.azquotes.com%2Fquote%2F602639&psig=AOvVaw2qqBhfBCAJAQ-4tzKhumlb&ust=1738478719508000&source=images&cd=vfe&opi=89978449&ved=0CBcQjhxqFwoTCLiF_qXwoYsDFQAAAAAdAAAAABAE))

WizT: The all in one platform to help you track your tasks. It's:
- Text-Based
- Simple to learn
- ~QUICK~ _VERY_ QUICK to use

All you need to do is

1. Download it 
2. Double-click it
3. Add your tasks
4. Let it manage your tasks for you! 🥳 

And guess what? It is **FREEEEEE!**

Features:

- [x] Managing Tasks
- [ ] Managing Deadlines (Stay Tuned!)
- [ ] Reminders (Stay Tuned!)

If you are a Java Programmer, you can use it to practice Java too. Here's the `main` method
```java
public class Main {
    }
}

### CODE REVIEW

File: gradle/wrapper/gradle-wrapper.jar
Comment (kumar2215) [inline gradle/wrapper/gradle-wrapper.jar:1]: Accidentally committed .jar file?

File: src/main/java/wizt/command/Command.java [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ *  Represents any commands that user inputs into program
+ */
+public class Command {
+    public Command() {
+
+    }
+
+    /*public void execute(TaskList tasks, Ui ui, Storage storage) throws WizTException {
```
kumar2215: Did you perhaps forget to remove this before committing?

src/main/java/wizt/command/Command.java
```diff
@@ -0,0 +1,31 @@
...
+
+    }*/
+
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws WizTException {
```
kumar2215, line 29: If there is no implementation meant here, perhaps it should be labelled abstract?

src/main/java/wizt/command/Command.java
```diff
@@ -0,0 +1,31 @@
...
+import wizt.task.TaskList;
+import wizt.ui.Ui;
+import wizt.ui.WizTException;
+
+
+
+/**
+ *  Represents any commands that user inputs into program
+ */
+public class Command {
```
kumar2215, line 16: Perhaps you meant this class to be abstract if you were following the guidelines?

image

src/main/java/wizt/ui/WizT.java
```diff
@@ -0,0 +1,117 @@
...
+
+        }
+    }
+
+    /**
+     * Write to hard disk
+     * @param filename  file name of the tasklist
+     * @param al Arraylist of all the tasks
+     */
+    public static void writeToFile(String filename, ArrayList<Task> al) {
```
kumar2215: Perhaps it makes more sense to put this functionality in the storage class?

src/main/java/wizt/ui/WizT.java
```diff
@@ -0,0 +1,117 @@
...
+
+
+
+    }
+
+    /**
+     * The main program execution
+     * @throws WizTException
+     */
+    public void run() throws WizTException {
```
kumar2215, line 46: Why does the run method throw `WizTException` if none of the code outside the try block throws it?

src/main/java/wizt/command/Command.java
```diff
@@ -0,0 +1,31 @@
...
+    public Command() {
```
kumar2215, line 17: I don't think an explicit empty constructor is necessary here since Java automatically provides one. Perhaps consider one that initializes the input to the command? This way the child command classes don't have to keep an empty constructor too.

src/main/java/wizt/ui/WizT.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            fw.close();
+        } catch (IOException e) {
+            System.out.println(e.getMessage());
+
+        }
+    }
+
+    public static void main(String[] args) throws WizTException {
```
kumar2215: I think main methods in general should not throw an exception. In this case, as stated for the run() method, there isn't a need for it here too.

src/main/java/wizt/ui/WizTException.java
kumar2215, line 1: Shouldn't this be in its own exceptions package and not the ui one?

src/main/java/wizt/parser/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+ *  Represents the different command the application recognises
+ */
+public class Parser {
+    /**
+     * Identify which command user has inputted
+     * @param fullCommand
+     * @return the corresponding command type
+     */
+    public static Command parse(String fullCommand) {
+        if (fullCommand.equals("list")) {
```
kumar2215, line 21: Perhaps you could restructure this? `else if` is a thing in Java.

src/main/java/wizt/parser/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+            return new ListCommand();
+        } else {
+            if (fullCommand.contains("unmark")) {
```
kumar2215: Is containing the right keyword a sufficient condition for it to be in the right format of a command? Or do the Command classes deal with cases where it contains the keyword but not in the right format (Eg. "pls unmark 3")?

src/main/java/wizt/command/UpdateCommand.java
```diff
@@ -0,0 +1,60 @@
...
+            int no = Integer.parseInt(split[1]);
+
+            al.get(no - 1).unmarkAsDone();
+            response.append("-------------------------------------")
+                    .append("Ok, I've marked this task as not done yet:")
+                    .append(al.get(no - 1).toString())
+                    .append("-------------------------------------");
+        } else {
+
+            if (input1.contains("mark")) {
```
kumar2215: Maybe consider separating this into 2 command classes: MarkCommand and UnmarkCommand?

src/main/java/wizt/command/AddCommand.java
```diff
@@ -0,0 +1,97 @@
...
+        StringBuilder response = new StringBuilder();
+        if (input1.contains("todo")) {
+
+            String substr = input1.substring("todo".length());
+            if (substr.isEmpty()) {
+                throw new WizTException("Please enter a description!");
+            }
+            Task t = new Todo(substr);
+            al.add(t);
+            response.append("\n -------------------------------------")
```
kumar2215: Some of these lines being appended here are being repeated in each of the different task types. Perhaps consider putting the string representations of each task type in their individual subclasses and then have a generic way of creating the response using them.

src/main/java/wizt/command/AddCommand.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Add corresponding task to tasklist based on command.
+     * @param tasks
+     * @param ui
+     * @param storage
+     * @throws WizTException
+     */
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws WizTException {
+        ArrayList<Task> al = tasks.getTasksList();
```
kumar2215: Variable name `al` doesn't seem to be meaningful here. Perhaps it could be renamed to taskList or not be created at all?

src/main/java/wizt/command/AddCommand.java
```diff
@@ -0,0 +1,97 @@
...
+            String substr = input1.substring("deadline".length());
+            if (substr.isEmpty()) {
+                throw new WizTException("Please enter a deadline value!");
+            }
+            String[] as = substr.split(" /by ");
+
+
+            //format dd/MM/YYYY HHmm
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+            LocalDateTime dt = LocalDateTime.parse(as[1], formatter);
```
kumar2215, line 98: A `DateTimeParseException` can be thrown here but doesn't seem to be handled.

src/main/java/wizt/command/AddCommand.java
```diff
@@ -0,0 +1,97 @@
...
+                    .append("\n Now you have " + al.size() + " in the list.")
+                    .append("\n -------------------------------------");
+        } else {
+            if (input1.contains("event")) {
+
+                String substr = input1.substring("event".length());
+                if (substr.isEmpty()) {
+                    throw new WizTException("Please enter a time period!");
+                }
+                String[] as = substr.split(" /from");
```
kumar2215: Perhaps you can consider the case where the string is not in the right format?

src/main/java/wizt/ui/WizT.java
```diff
@@ -0,0 +1,117 @@
...
+        new WizT("wizt.txt").run();
+
+    }
+
+
+    public String getResponse(String fullCommand) throws WizTException {
```
kumar2215, line 94: The only method call for this method is in this file and is being commented out. Perhaps you can consider removing it if it's not being used?

src/main/java/wizt/ui/WizT.java
```diff
@@ -0,0 +1,117 @@
...
+
+        boolean isExit = false;
+        while (!isExit) {
+
+            try {
+                 String fullCommand = ui.readCommand();
+               // getResponse(fullCommand);
+
+                Command c = Parser.parse(fullCommand);
+                c.execute(tasks, ui, storage);
```
kumar2215, line 57: Each of the command classes have String as the return type of their execute method but that is not being used here. Perhaps consider making them have a return type of void then?

src/main/java/wizt/command/UpdateCommand.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param storage
+     */
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
+        ArrayList<Task> al = tasks.getTasksList();
+        StringBuilder response = new StringBuilder();
+        if (input1.contains("unmark")) {
+            String[]split = input1.split(" ");
+
+            int no = Integer.parseInt(split[1]);
```
kumar2215: `NumberFormatException` can be thrown here if the format of the string is not what is expected to be. Perhaps consider handling it.

src/main/java/wizt/command/UpdateCommand.java
```diff
@@ -0,0 +1,60 @@
...
+            response.append("-------------------------------------")
+                    .append("Ok, I've marked this task as not done yet:")
+                    .append(al.get(no - 1).toString())
+                    .append("-------------------------------------");
+        } else {
+
+            if (input1.contains("mark")) {
+                String[]split = input1.split(" ");
+                int no = Integer.parseInt(split[1]);
+                al.get(no - 1).markAsDone();
```
kumar2215: Shouldn't the TaskList class be responsible for marking, unmarking, deleting and adding of tasks? I don't think it's a good idea to directly access the task here in any of the command classes.

kumar2215 (commented):
I believe the quality of code can be improved here. There are a lot of places where exceptions can be thrown but are not handled.

src/main/java/wizt/task/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ *  Represents the Deadline Task
+ */
+public class Deadline extends Task {
+
+    protected LocalDateTime by;
```
xa0412, line 11: Would it be better to make LocalDateTime private and accessed via a getter to enforce encapsulation?

src/main/java/wizt/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Create a file if does not exist
+     * @param filename
+     */
+    public Storage(String filename) {
+        this.filename = filename;
+        File f = new File(filename);
+        try {
+            if (f.createNewFile()) {
+                System.out.println("File dont exist!, created a new file wizt.txt!");
```
xa0412, line 32: (File does not exist!) rather then (File dont exist!)?

src/main/java/wizt/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+     * Load the tasks from file into Arraylist
+     * @return all tasks in the task list
+     * @throws FileNotFoundException
+     */
+    public ArrayList<Task> load() throws FileNotFoundException {
+
+
+        File filename = new File(this.filename);
+        Scanner s = new Scanner(filename);
+        ArrayList<Task> al = new ArrayList<>();
```
xa0412: Use meaningful variable name?
s -> scanner
al -> tasks

src/main/java/wizt/ui/Main.java
```diff
@@ -0,0 +1,36 @@
...
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private WizT wizt = new WizT("wizt.txt");
```
xa0412: Since "wizt.txt" is the default file name, store it in a constant?
private static final String FILENAME = "wizt.txt";

src/main/java/wizt/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+
+        while (s.hasNext()) {
+            String line = s.nextLine();
+
+            if (line.contains("from")) {
+                line = line.substring(7);
+
+                String[] as = line.split("\\(from:");
+                String[] as2 = as[1].split("to:");
```
xa0412: Same: Use meaningful variable name?

xa0412 (commented):
Looks good, just some advice from me.

PR #85, opened 2025-01-25
[wzhua02] iP

wzhua02:
- Keeps track of stuff
     - And more.
---
1. Link to `Google` [here](https://www.google.com)
2. Or use other search engines
3. Wow emojis :smiley_cat:

Make sure to:
- [ ] Do your iP
- [ ] Meet with your team for tP
- [x] Do your weekly quiz

```
public void sayGdWork() {
    System.out.println("Good work");
}
```


src/main/java/baymax/Baymax.java
```diff
@@ -0,0 +1,51 @@
...
+ * Handles user interactions and task management.
+ */
+public class Baymax {
+    private static final Path FILE_PATH = Paths.get("data", "tasks.txt");
+    private final Storage storage;
+    private final TaskList tasks;
+    private final Ui ui;
+
+    /**
+     * Constructs a Baymax instance.
```
clx3210, line 29: Shall we add a newline here between the main description and the `@param`?

src/main/java/baymax/Baymax.java
```diff
@@ -0,0 +1,51 @@
...
+        String input = ui.getInput();
+        while (!input.startsWith("bye")) {
+            Parser.parse(input, ui, tasks, storage);
+            input = ui.getInput();
+        }
+        ui.reply("Goodbye! *slowly deflates*");
+    }
+
+    /**
+     * The main method to start the chatbot.
```
clx3210: Same comment as above.

src/main/java/baymax/io/Storage.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Loads the list of tasks from the file.
+     * Each line in the file represents a single task.
+     *
+     * @return an ArrayList containing the tasks read from the file
+     */
+    public ArrayList<String> load() {
+        ArrayList<String> returnList = new ArrayList<>();
```
clx3210: Shall we rename this to a more meaningful noun? For example, how about `rawDataObjects`?

src/main/java/baymax/io/Ui.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.PrintStream;
+import java.util.Scanner;
+
+/**
+ * Represents a user interface (UI) for interacting with the user via standard input and output.
+ * This class is responsible for reading user input and formatting output with indentation.
+ */
+public class Ui {
+    private static final String INDENT = "    ";
+    private final Scanner scan;
```
clx3210: Shall we rename to the `scanner` noun instead?

src/main/java/baymax/task/Deadline.java
```diff
@@ -0,0 +1,71 @@
+package baymax.task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a deadline task with a specific due date and time.
+ */
+public class Deadline extends Task {
```
clx3210, line 9: I like the naming convention used for this class, It is very clear.

src/main/java/baymax/task/Task.java
```diff
@@ -0,0 +1,66 @@
...
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Marks or unmarks the task as completed.
+     *
+     * @param mark {@code true} to mark as done, {@code false} to unmark.
+     */
+    public void marker(boolean mark) {
```
clx3210: Shall we rename `marker` to a verb? For example:
```java
public void markDoneStatus(boolean isDone) {
...
}
```

src/main/java/baymax/task/TaskList.java
```diff
@@ -0,0 +1,147 @@
...
+            return foundTasks;
+        }
+    }
+
+    /**
+     * Returns the number of tasks in the list.
+     *
+     * @return The size of the task list.
+     */
+    public int size() {
```
clx3210: How about renaming this to `getSize`?

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+            case "unmark" -> {
+                if (args.length < 2) {
+                    throw new BaymaxException("Do let me know which task to mark/unmark.");
+                }
+                int idx = Integer.parseInt(args[1]) - 1;
+                if (idx < 0 || idx >= tasks.size()) {
+                    throw new BaymaxException("I do not know which task you are referring to.");
+                }
+                Task theTask = tasks.getTask(idx);
```
clx3210: How about `taskToUnmark`? This may help improve readability.

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,196 @@
...
+            case "delete" -> {
+                String[] parts = input.split(" ");
+                if (parts.length < 2) {
+                    throw new BaymaxException("Do let me know which task to mark/unmark.");
+                }
+                int idx = Integer.parseInt(parts[1]) - 1;
+                if (idx < 0 || idx >= tasks.size()) {
+                    throw new BaymaxException("I do not know which task you are referring to.");
+                }
+                Task theTask = tasks.getTask(idx);
```
clx3210: Same comment as above

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,196 @@
...
+                        "Now you have " + tasks.size() + " tasks in the list.");
+            }
+            case "bye" -> {
+                // Do nothing
+            }
+            case "find" -> {
+                if (args.length < 2) {
+                    throw new BaymaxException("Let me know what task you would like to find.");
+                }
+                ArrayList<Task> foundTasks = tasks.findTask(args[1]);
```
clx3210: I like the variable naming here! The variable naming here provides a clear idea of what the code is doing.

src/test/java/baymax/io/StorageTest.java
```diff
@@ -0,0 +1,44 @@
...
+        storage = new Storage(testFile);
+    }
+
+    @Test
+    void testFileCreation() {
+        assertTrue(Files.exists(testFile));
+    }
+
+    @Test
+    void testSaveAndLoad() {
```
clx3210: Shall we rename the methods here to use the naming convention for testing methods? Example: `saveAndLoad_multipleTasks_success()`

src/test/java/baymax/io/UiTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @BeforeEach
+    void setUp() {
+        outputStream = new ByteArrayOutputStream();
+        PrintStream printStream = new PrintStream(outputStream);
+        inputStream = new ByteArrayInputStream("test input\n".getBytes());
+        ui = new Ui(inputStream, printStream);
+    }
+
+    @Test
+    void testGetInput() {
```
clx3210: Same comment as above for all the unit test methods

clx3210 (commented):
Looks good! 👍 
Just some minor comments on code quality naming guidelines.

src/main/java/baymax/task/TaskList.java
```diff
@@ -0,0 +1,168 @@
...
+
+import baymax.exception.BaymaxException;
+import baymax.util.Parser;
+import baymax.util.Storage;
+
+/**
+ * Manages a list of tasks, allowing addition, removal, retrieval, and persistence.
+ */
+public class TaskList {
+    private final ArrayList<Task> taskList;
```
zihung20, line 16: Shouldn't this to be named as plural?`private final ArrayList<Task> taskLists `

src/main/java/baymax/task/Deadline.java
```diff
@@ -0,0 +1,71 @@
+package baymax.task;
```
zihung20, line 1: I like that you put all related classes into a package

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public class Parser {
+    private static final String INDENT = "    ";
+    /**
+     * Parses the user input and executes the appropriate command.
+     *
+     * @param input   The user input string.
+     * @param tasks   The TaskList containing user tasks.
+     * @param storage The Storage handler for saving and loading tasks.
+     * @return A String that Baymax will reply to the user based on the input.
```
zihung20: Isn't this should be An array of String?

zihung20 (commented):
Overall, you strictly followed the coding standard, and there aren't many coding standard issues in your files. Good job keeping the code standardized.

src/main/java/baymax/task/Event.java
```diff
@@ -0,0 +1,87 @@
...
+     * Converts the event task into a formatted string for saving.
+     *
+     * @return A formatted string representing the event for storage.
+     */
+    @Override
+    public String toSaveFormat() {
+        return "E | " + (isDone ? "1" : "0") + " | " + getDescription()
+                + " | " + this.fromDate.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"))
+                + " | " + this.toDate.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"));
+    }
```
iuhiah:
```suggestion
    @Override
    public String toString() {
        String typeStr = "[" + this.type + "]";
        String fromToStr = " (from: " + getFromDate() + ") (to: " + getToDate() + ")";
        return typeStr + super.toString() + fromToStr;
    }

    /**
     * Converts the event task into a formatted string for saving.
     *
     * @return A formatted string representing the event for storage.
     */
    @Override
    public String toSaveFormat() {
        return "E | " + super.toSaveFormat()
                + " | " + this.fromDate.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"))
                + " | " + this.toDate.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"));
    }
```

Perhaps move the common substrings to the superclass for better abstraction and code management!

src/main/java/baymax/task/TaskList.java
iuhiah: I think `taskList` would be the better option! This class would only represent one task list that contains multiple tasks, so it would either be `taskList` or `tasks`!

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,174 @@
...
+                if (args.length < 2) {
+                    throw new BaymaxException("Do let me know which task to mark/unmark.");
+                }
+                int idx = Integer.parseInt(args[1]) - 1;
+                if (idx < 0 || idx >= tasks.size()) {
+                    throw new BaymaxException("I do not know which task you are referring to.");
+                }
+                Task theTask = tasks.getTask(idx);
+                theTask.marker(true);
+                tasks.save(storage);
```
iuhiah: Not a coding quality standard comment(?), but I would think about only asking the Parser class to parse data and nothing more (e.g. adding the task, saving tasks to storage)! This will help in isolating the class for testing in case there are other issues with the TaskList and Storage classes :)

src/main/java/baymax/util/Parser.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param dateTimeStr The date-time string to be parsed.
+     * @return A LocalDateTime object representing the parsed date and time.
+     * @throws BaymaxException If the date-time string does not match any valid format.
+     */
+    public static LocalDateTime parseDateTime(String dateTimeStr) throws BaymaxException {
+        String[] patterns = {
+            "yyyy-MM-dd HH:mm", //e.g. 2025-01-27 12:30
+            "dd/MM/yyyy HH:mm", //e.g. 27/01/2025 12:30
+            "yyyy MM dd HH:mm", //e.g. 2025 01 27 12:30
+        };
```
iuhiah: Consider setting this as a static class variable! Else a new DateTimeFormatter object will be created whenever this method is called.

iuhiah (commented):
Project is well organised into packages! Some nits have already been pointed out by other reviewers (e,g, getter naming convention). I would think about dereferencing class if possible (as mentioned in the Parser class) as it seems quite tough to carry out isolated unit tests with the current structure.

P.S. Baymax concept is super cute!

src/main/java/baymax/task/TaskList.java
wzhua02: good point, will change this.

src/main/java/baymax/task/Task.java
wzhua02: good point, have changed the method to just be markTask() and unmarkTask() instead.

src/main/java/baymax/task/Event.java
wzhua02: Good suggestion, I think that looks much cleaner, thanks!

src/main/java/baymax/util/Parser.java
wzhua02: True good catch! Planning to change this into command class anyway, will be abstracting command functionality into individual classes.

src/main/java/baymax/util/Parser.java
wzhua02: Yup have been planning to create classes for each command to separate their functionality from this method, will be doing this in the next few commits, but thanks for pointing it out.

src/main/java/baymax/util/Parser.java
wzhua02: Yup will be adding a method unmarkTask in the next few commits

src/main/java/baymax/util/Parser.java
wzhua02: Good point!

PR #87, opened 2025-01-26
[Teo Zi Hui, Shelia] iP

shelialia:
# Task Assistant Rose
> You don't have to see the whole staircase, just take the first step" — Martin Luther King, Jr. [View Source](https://www.goodreads.com/quotes/1063592-you-don-t-have-to-see-the-whole-staircase-just-take)

Here's how Rose makes _your_ day better
* Keeps track of your basic tasks, deadlines, and events
* Easy search up for tasks
* Easy addition and deletion
* and more upcoming!! 💯 

But what's so special about Rose?
1. Run here directly from your laptop
2. Get rid of websites, just run her locally
3. Your personal assistant

And it is *FREE*!

I used Java to build the app. Here is the main entry point of the application.
```java
public class Main {
    }
}
```
Why not you try it for yourself?


src/main/java/rose/ui/Event.java
```diff
@@ -0,0 +1,52 @@
...
+        this.to = to;
+    }
+
+    /**
+     * Converts this event into a file format string suitable for saving.
+     *
+     * @return A string representing this event in file format.
+     */
+    @Override
+    public String toFileFormat() {
```
tim0tay: I like that the naming of this method (and for the other tasks too) are clearly understandable!

src/main/java/rose/ui/Rose.java
```diff
@@ -0,0 +1,126 @@
...
+
+        Scanner scanner = new Scanner(System.in);
+
+        while (true) {
+            String input = scanner.nextLine();
+
+            try {
+                Command command = Parser.parse(input);
+
+                switch (command.getCommand().toLowerCase()) {
```
tim0tay: Would it be better to encapsulate this parsing in the parser class?

src/main/java/rose/ui/Deadline.java
```diff
@@ -0,0 +1,44 @@
+package rose.ui;
```
tim0tay: Perhaps these tasks (Deadline, Event, Todo) can be in a Task package instead?

src/main/java/rose/ui/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+     * @return A {@link Task} object corresponding to the line.
+     * @throws RoseException If the line format is invalid or the task type is unknown.
+     */
+    private Task parseTask(String line) throws RoseException {
+        String[] parts = line.split(" \\| ");
+        String taskType = parts[0];
+        boolean isDone = parts[1].equals("1");
+
+        switch (taskType) {
+            case "T":
```
tim0tay: Mayhaps 'tis better to not indent cases!

src/main/java/rose/ui/Ui.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Constructs a new {@code Ui} object.
+     */
+    public Ui() {}
+
+    /**
+     * Displays the greeting message when the program starts.
+     */
+    public void greet() {
+        System.out.println("____________________________________________________________");
```
tim0tay: I love how you separated the horizontal line into two lines! 👍

tim0tay (commented):
Shelia this looks amazing! Can tell that you've put in a lot of effort and diligence into ensuring that your code is of good quality :> 🥇

PR #91, opened 2025-01-26
[Heng E Chun] iP

juniper285:
Loading___

### Gigi

> “The brain is like a muscle. When it is in use, we feel very good. Understanding is joyous.”
> – Carl Sagan ([source](https://www.nytimes.com/1979/06/10/archives/outer-space-and-inner-space-sagan.html?smid=url-share))

Look, you’ve got too much rattling around in that tiny mortal brain of yours. I, **the great Gigi**, exist so you don’t have to waste your precious energy remembering stuff. 

> [!CAUTION]
> Here’s why you’ll love it:
- It’s text-based (_simple, no nonsense_).
- It’s easy to learn (_even for you_).
- It’s ~~FAST~~ SUPER FAST! (_like me when someone tries to pick me up_).

## What do you have to do?
Simple. Even a human can manage this:

1. Download it from [here](https://nus-cs2103-ay2425s2.github.io/website/admin/ip-w4.html).
2. Double-click it (seriously, just two clicks).
3. Add your tasks (or don't, but don’t blame me when you forget everything).
4. Let it handle the rest while you do... whatever it is you do.

Features:
- [ ]  Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)

```
public class Main {
    }
}
```

There. Now go forth, download Gigi, and stop forgetting things. You’re welcome. 😼🔥


src/main/java/gigi/main/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+    );
+    protected static final List<DateTimeFormatter> DATE_FORMATTERS = List.of(
+            DateTimeFormatter.ofPattern("MM/dd/yyyy"),
+            DateTimeFormatter.ofPattern("dd/MM/yyyy"),
+            DateTimeFormatter.ofPattern("yyyy-MM-dd"),
+            DateTimeFormatter.ofPattern("d MMM yyyy"),
+            DateTimeFormatter.ofPattern("d MMMM yyyy")
+    );
+
+    public static Command parse(String input) throws GigiException {
```
lucasisnotcool: Perhaps a header comment here for this public method?

Per coding standard:
"Write descriptive header comments for all public classes/methods.

You MUST write header comments for all classes, public methods. But they can be omitted for the following cases:
i. Getters/setters
ii. When overriding methods (provided the parent method's Javadoc applies exactly as is to the overridden method)"

src/main/java/gigi/commands/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+
+import gigi.exceptions.GigiException;
+import gigi.storage.Storage;
+import gigi.tasks.Task;
+import gigi.tasks.Tasklist;
+import gigi.ui.Ui;
+
+public class DeleteCommand extends Command {
+    public static final String COMMAND_WORD = "delete";
+    public final int taskIndex;
```
lucasisnotcool: Any reason this int taskIndex is public? 

Per coding standard:
"Class variables should never be declared public unless the class is a data class with no behavior. This rule does not apply to constants."

src/main/java/gigi/commands/Command.java
```diff
@@ -0,0 +1,31 @@
...
+import gigi.exceptions.GigiException;
+import gigi.storage.Storage;
+import gigi.tasks.Tasklist;
+import gigi.ui.Ui;
+
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+
+public abstract class Command {
+    protected static final List<DateTimeFormatter> FORMATTERS = List.of(
```
lucasisnotcool: Love that you support SO MANY formats! However, I noticed that Formatters is found in both Command.java and Parser.java. Is there a reason for this?

src/main/java/gigi/commands/UnmarkCommand.java
```diff
@@ -0,0 +1,32 @@
...
+    public final int taskIndex;
+
+    public UnmarkCommand(int i) {
+        this.taskIndex = i - 1;
+    }
+
+    @Override
+    public void execute(Tasklist tasks, Ui ui, Storage storage) throws GigiException {
+        if (taskIndex < 0 || taskIndex >= tasks.getSize()) {
+            throw new GigiException("MEOW! You don't have that many tasks.");
```
lucasisnotcool, line 43: Cool error message 😄

src/main/java/gigi/tasks/Deadlines.java
```diff
@@ -0,0 +1,44 @@
...
+        super(taskName);
+        this.deadline = deadline;
+        this.isComplete = isComplete;
+    }
+
+    public String getStatusIcon() {
+        return "[D]" + super.getStatusIcon(); // mark done task with X
+    }
+
+    void markDone(int index) {
```
lucasisnotcool, line 66: Maybe these marking functions are not neccessary in children classes (deadline, event) to task? Since execution will fallback to definition from parent class 🤔

src/main/java/gigi/tasks/ToDos.java
```diff
@@ -0,0 +1,37 @@
...
+        this.isComplete = false;
+    }
+
+    public ToDos(String taskName, boolean isComplete) {
+        super(taskName);
+        this.isComplete = isComplete;
+    }
+
+    public String getStatusIcon() {
+        return "[T]" + super.getStatusIcon(); // mark done task with X
```
lucasisnotcool: perhaps 'T' can be stored as constant so it does not need to be repeated?

src/main/java/gigi/tasks/ToDos.java
```diff
@@ -0,0 +1,37 @@
+package gigi.tasks;
+
+public class ToDos extends Task{
+    protected String taskName;
+    protected boolean isComplete;
```
lucasisnotcool: Instead of passing the parent's internals around, maybe we can implement this using setters and getters? 🤔

src/main/java/gigi/storage/Storage.java
```diff
@@ -0,0 +1,63 @@
+package gigi.storage;
+
+import gigi.exceptions.GigiException;
+import gigi.main.Parser;
+import gigi.tasks.Deadlines;
+import gigi.tasks.Events;
+import gigi.tasks.Task;
+import gigi.tasks.ToDos;
+
+import java.io.*;
```
lucasisnotcool: perhaps we should import the classes we need explicitly, looks like you only need BufferredReader and PrintWriter? Noticed this issue in Parser.java as well!

Per coding standard: 
"Imported classes should always be listed explicitly.

Good
```java
import java.util.List;
import java.util.ArrayList;
import java.util.HashSet;
```

Bad
```java
import java.util.*;
```
"

src/main/java/gigi/main/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+        String[] deadlineDetails = details.split(" /by ", 2);
+        if (deadlineDetails.length < 2 || deadlineDetails[0].isBlank() || deadlineDetails[1].isBlank()) {
+            throw new GigiException("MEOW!!! The description and date of a deadline cannot be empty.");
+        }
+        return new DeadlineCommand(deadlineDetails[0], parseDateTime(deadlineDetails[1]));
+    }
+
+    private static Command startEvent(String details) throws GigiException {
+        String[] eventDetails = details.split(" /from | /by ", 3);
```
lucasisnotcool: wow! I didn't know you could do regex like this. will steal for my implementation 🤩

lucasisnotcool (commented):
LGTM. Some minor changes that do not affect functionality ⭐

src/main/java/gigi/Gigi.java
```diff
@@ -0,0 +1,48 @@
...
+            ui.showLoadingError();
+            tasks = new Tasklist();
+        }
+    }
+
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
```
Sean-G-Han, line 54: I like how you are using OOp principles here

src/main/java/gigi/Gigi.java
```diff
@@ -0,0 +1,48 @@
...
+        new Gigi(FILE_PATH).run();
+    }
+
+    public Gigi(String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new Tasklist();
+            tasks.getTasks(storage);
+        } catch (GigiException e) {
```
Sean-G-Han, line 33: Maybe having subclasses of GigiException for more specific cases will be better, e.g. GigiEmptyString...

PR #96, opened 2025-01-26
[feliciahmq] iP

feliciahmq:
### Claudia

> "Your mind is for having ideas, not holding them." - David Allen ([source](https://dansilvestre.com/productivity-quotes/))

Claudia frees your mind of having to remember things you need to do. It's, 

- text-based
- easy to learn
- ~FAST~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/feliciahmq/ip/releases).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you  😉

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---

``` java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/claudia/ui/Claudia.java [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+ * The main class for Claudia chatbot.
+ * It communicates with Ui, Storage and TaskList to handle user interactions.
+ */
+public class Claudia {
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Constructs a Friday chatbot, initializing user interface, storage and TaskList.
```
NightlyAffair: I think meant to say constructs a Claudia chatbot?

src/main/java/claudia/ui/Claudia.java
```diff
@@ -0,0 +1,69 @@
...
+        } catch (ClaudiaException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Runs Claudia chatbot, reading and executing user command
+     * till exit command is given by user.
+     */
```
NightlyAffair, line 42: I think this is a great structure for the chatbot, keeps everything easy to track and trace 👍

src/main/java/claudia/ui/Ui.java
```diff
@@ -0,0 +1,173 @@
...
+     * Displays a message showing the information of the Event task added.
+     *
+     * @param tasks The current list of tasks.
+     * @param event The Event task that was added.
+     */
+    public void showEvent(TaskList tasks, Event event) {
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("  " + event.toString());
+        System.out.printf(" Now you have %d tasks in the list.\n", tasks.size());
+    }
```
NightlyAffair: Why not have one method as below to handle all 3 cases by taking in a task?

```
public void showEvent(TaskList tasks, Task task) {
        System.out.println(" Got it. I've added this task:");
        System.out.println("  " + task.toString());
        System.out.printf(" Now you have %d tasks in the list.\n", tasks.size());
    }
```

NightlyAffair (commented):
I think very clean code with methods in logical places allowing for very structured packages and classes.

src/main/java/claudia/ui/Claudia.java
feliciahmq: Agreed.

src/main/java/claudia/ui/Ui.java
feliciahmq: Agreed, I will add a new method to reduce repetition of code.

src/main/java/claudia/parser/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+        FIND,
+        UNKNOWN;
+
+        /**
+         * Converts a string into a CommandType.
+         *
+         * @param command The command string.
+         * @return The valid CommandType or Unknown if invalid.
+         */
+        public static CommandType fromString(String command) {
```
Golddirio, line 45: Maybe the naming can be clearer?

src/main/java/claudia/task/Deadline.java
```diff
@@ -0,0 +1,65 @@
...
+        super(description);
+        this.by = by;
+    }
+
+    /**
+     * Returns a formatted string of Deadline task to save to storage file.
+     *
+     * @return The file format representation of the Deadline task.
+     */
+    public String fileFormat() {
```
Golddirio, line 44: Maybe the naming can use a verb instead?

feliciahmq: Sounds good! Thanks for feedback :)

src/main/java/claudia/parser/Parser.java
feliciahmq: Agreed!

PR #102, opened 2025-01-26
[alexandtheoh] iP

alexandtheoh:
# *Alex Pull Request*
- this is a bullet list 🤟🏻
- this 2
1. [this is a numbered list](https://youtube.com/shorts/1c5pI9nGtnc?si=P1GWV9ToPbt0yzLv)
2. this 2
To call the function, use the command `new CodeBlock()` in your script.
```java
public CodeBlock(String dateString) {
    try {
        date = LocalDate.parse(dateString);
        isDate = true;
    } catch (Exception e) {
        this.dateString = dateString;
        isDate = false;
    }
}
```
- [x] this is done
- [ ] this is not done
> [!NOTE]
> Useful information that users should know, even when skimming content.


Clifong (commented):
The chain of append looks verbose. Suggestion to just use normal string concatenation since performance isn't a concern here.

Clifong (commented):
IMO, the System.out.println preferably should be one sentence/variable per line
E.g
System.out.println("I am bored. Let's play a game called: "
 /+ gameName
 /+ ". Whatever."

I think this is nicer because otherwise you have to scroll to the right.

Clifong (commented):
What is qq?

Clifong (commented):
I like the choice of having a TaskDate object

src/main/java/alexis/ui/Ui.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public void start() {
+        // Create a Scanner object to read input from the command line
+        Scanner scanner = new Scanner(System.in);
+
+        // intro line
+        System.out.println(line + "Hello! I'm Alexis\n"
+                + "What can I do for you?\n" + line);
+
+        // Infinite loop to continuously echo input
```
zechary28: you should update the comments, echo was like level 2

src/main/java/alexis/ui/Ui.java
```diff
@@ -0,0 +1,164 @@
...
+        while (scanner.hasNextLine()) {
+            // Prompt the user for input
+            String input = scanner.nextLine();
+
+            // Exit condition
+            if (ActionsEnum.bye.name().equalsIgnoreCase(input)) {
+                storage.save();
+                System.out.println(line + "Bye. Hope to see you again soon!\n" + line);
+                break;
+                // display list condition
```
zechary28: comment should follow indentation of respective code, i think in this case should be 4 spaces back to follow code below

src/main/java/alexis/ui/Ui.java
```diff
@@ -0,0 +1,164 @@
...
+                } else if (words.length == 2 && isInteger(words[1])) { // case of mark/unmark/delete
+                    if (ActionsEnum.mark.name().equalsIgnoreCase(words[0])) {
+                        int pos = Integer.parseInt(words[1]);
+                        try {
+                            Task task = tasksList.markTask(pos);
+                            System.out.println(line + "Nice! I've marked this task as done:\n"
+                                    + task.getStatusIcon() + " " + task.getDescription() + "\n" + line);
+                        } catch (InvalidPosException e) {
+                            System.out.println(Ui.invalid);
+                        }
```
zechary28: maybe u can define this as a function, so its shorter and more readable. same with unmark and delete below

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public String getDate() {
+        return "(by: " + by.toString() + ")";
+    }
+
+    @Override
+    public String toString() {
+        return type + this.getStatusIcon() + " " + description + " (by: " + by.toString() + ")";
```
zechary28, line 32: you could call getType() and getDate() here

src/main/java/tasks/TasksList.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Deletes a task from the list
+     *
+     * @param pos is the position of the task to be deleted, InvalidPosException is thrown if pos is out of bounds
+     *
+     * @return is the task which was deleted
+     */
+    public Task deleteTask(int pos) throws InvalidPosException {
```
zechary28, line 37: good catch for invalid input above array size and <= 0

zechary28 (commented):
overall pretty good, just a few nits. I would structure stuff a bit differently but yours seems to do the job

PR #113, opened 2025-01-26
[mihirniyogi] iP

mihirniyogi:
# Bob

_Is that a bird? Is that a plane?_ 

No, it's 🔥<mark>BOB</mark>🔥 

- You may ask... why 'Bob'? 
- Why not some fancy name like 'Duke Royalty'?
- or 'iPhone XS Max Ultra+ Pro'?

because I believe in 

> Simplicity™️  

Much wow! So🆒COOL🆒! Bob The Task Manager is here to help make your task management 😎 SIMPLE 😎. 
Features include:

1. add todo, event or deadline
2. mark / unmark tasks
3. list all tasks
4. remove tasks
5. find tasks

Almost as simple as a notepad, like:
- [x] apples
- [x] bananas
- [ ] cherries

But almost as powerful as a full-on desktop application! For example, 'list' brings up all your tasks. 

&emsp;&emsp;Here are your tasks:
&emsp;&emsp;1. [E][ ] project meeting (from: Mar 17 2025 2:00pm to: Mar 17 2025 6:00pm)
&emsp;&emsp;2. [T][X] read book
&emsp;&emsp;3. [T][ ] return book
&emsp;&emsp;4. [D][ ] assignment (by: Feb 3 2025 11:59pm)

One day, Bob will become King Bob👑, see [here](https://www.youtube.com/watch?v=TjAg-8qqR3g).

---

For the nerds out there who want to edit Bob's precious source code to add more commands 🙄, you should edit `Parser::parse` and add another class that inherits from `Command`, as shown below:

```java
public class Parser {
    public static Command parse(String userInput) throws WrongCommandException, IOException {
        if (userInput.equals("bye")) {
            return new ExitCommand();
        }

        if (userInput.equals("list")) {
            return new ListCommand();
        }

        if (userInput.startsWith("mark")) {
            int number = Integer.parseInt(userInput.split(" ")[1]);
            return new MarkCommand(number);
        }

        // Add if-blocks here to catch more commands //
        // ...
    }    
}
```


src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,46 @@
...
+     * <p>
+     * This method initializes the user interface, prints the welcome message,
+     * and enters an infinite loop to read and process user commands.
+     * It handles various exceptions, such as invalid commands and input errors,
+     * and provides feedback to the user through the {@link Ui} class.
+     * </p>
+     */
+    public static void main(String[] args) {
+
+        Ui ui = new Ui();
```
cyrolite: Consider declaring Ui as a final variable since it is not reassigned after initialization. 😄

src/main/java/bob/parser/Parser.java
```diff
@@ -0,0 +1,127 @@
...
+        if (userInput.equals("bye")) {
+            return new ExitCommand();
+        }
+
+        if (userInput.equals("list")) {
+            return new ListCommand();
+        }
+
+        if (userInput.startsWith("mark")) {
+            int number = Integer.parseInt(userInput.split(" ")[1]);
```
cyrolite: It would be good if there was a way to check if the  split array has at least 2 elements to avoid `ArrayIndexOutOfBoundsException`.

src/main/java/bob/command/Command.java
cyrolite, line 1: I like how informative and clear the JavaDoc comments are. Good job!

src/main/java/bob/storage/Storage.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public static void saveTasksToFile(List<Task> tasks) throws IOException {
+        String lines = Stream.concat(Stream.of(HEADER),
+                tasks
+                    .stream()
+                    .map(task -> (tasks.indexOf(task) + 1) + "," + task.toCsv()))
+                .collect(Collectors.joining("\n"));
+        Files.write(FILE_PATH, lines.getBytes(StandardCharsets.UTF_8), StandardOpenOption.TRUNCATE_EXISTING);
+    }
+}
```
cyrolite, line 141: LGTM, excellent use of lambda expressions to help you with your application. I also like how tidy and organised the code is. Good job!

src/main/java/bob/task/TaskList.java
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param number (1-indexed).
+     * @throws IOException if error during file IO.
+     * @throws IndexOutOfBoundsException if number is out of range.
+     */
+    public static void unmarkTask(int number) throws IOException, IndexOutOfBoundsException {
+        tasks.get(number - 1).unmark();
+        Storage.saveTasksToFile(tasks);
+    }
+}
```
cyrolite, line 114: These are some well-handled exception cases (e.g., IO errors, index out of bounds), coupled with appropriate messages. Well done!

src/main/java/bob/Bob.java
mihirniyogi: oh yea good point ! will do

src/main/java/bob/parser/Parser.java
mihirniyogi: ah ok good point actually, this goes to show that my testing also needs improvement to catch this error 🤣

src/main/java/bob/storage/Storage.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    /**
+     * Fetches tasks from the file and returns them as a list.
+     * Returns empty list if encountered any IO error.
+     *
+     * @return List of tasks.
+     */
+    public static List<Task> fetchTasksFromFile() {
```
Partillay, line 94: Perhaps the method name can be shortened to fetchTasks() instead?

Partillay (commented):
The import commands are arranged nicely and overall code is really neat!

src/main/java/bob/storage/Storage.java
mihirniyogi: good point👍

PR #115, opened 2025-01-26
[Wong An Wei] iP

AnWe11:
### **Luigi**

> “Your mind is for having ideas, not holding them.” – David Allen

Luigi helps to keep track of your tasks so you dont have to! 😉

Its features include,

- Adding / Deleting tasks
- Marking/ Unmarking tasks
- Local memory update
- Encouraging messages from Luigi

All you need to do is, 

1. Download it from [here](https://github.com/AnWe11/ip/releases/tag/Luigi-v2)
2. Double Click the file
3. Add your tasks
4. Let it manage for you

Other features **coming soon** include,

- _[ ] Reminders_
- _[ ] Prioritisation of tasks_

`
public class Main {
    }
}
`


src/main/java/commands/ByeCase.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Returns goodbye message and exits the program.
+ */
+public class ByeCase implements DefaultCase {
+
+    public ByeCase() { }
+
+    /**
+     * Returns goodbye message and exit the program.
+     */
```
chanjunlin, line 12: Nice use of Javadoc comments for the actions() method! it clearly states what the method does. You might want to specify that this method returns a string when called. 😊

src/main/java/commands/CommandsParser.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Returns corresponding methods based on respective command words.
+ */
+public class CommandsParser {
+
+
+    // List of command keywords the user can input that will trigger a method call.
+    enum Keywords {
+        BYE, LIST, MARK, UNMARK, TODO, DEADLINE, EVENT, DELETE, FIND, INVALID;
```
chanjunlin, line 15: I like how you used enum to manage command keywords. It makes your code cleaner and easier to manage! Would it be better to handle INVALID as part of error checking within the parsing logic 🤔?

src/main/java/commands/CommandsParser.java
```diff
@@ -0,0 +1,92 @@
...
+                break;
+            case INVALID:
+                InvalidCase invalidCase = new InvalidCase(input);
+                responseString = invalidCase.action();
+                break;
+            }
+        } catch (InvalidCommandException e) {
+            System.out.println(e.getMessage());
+            responseString = e.getMessage();
+        }
```
chanjunlin, line 87: The use of switch based on the keyword enum makes it easy to follow which command maps to which actions. You Perhaps adding a default case in your switch statement can handle any unforeseen scenarios gracefully?

src/main/java/commands/EventsCase.java
```diff
@@ -0,0 +1,67 @@
...
+            String toString = fromToString[1];
+            Events eventsTask = new Events(taskDescription, fromString, toString);
+            responseString = taskManager.addTask(eventsTask);
+            data.saveData(taskManager);
+        } catch (InvalidCommandException e) {
+            System.out.println(e.getMessage());
+//            System.out.println("_____________________________________________");
+//            System.out.println("Input format for events tasks should be\nevent <Task Description> /from <from date> /to <to date>");
+//            System.out.println("_____________________________________________");
+            taskDescription = "";
```
chanjunlin: I think this assignment might be unnecessary as you are not using taskDescription afterwards?

src/main/java/commands/FindCase.java
```diff
@@ -0,0 +1,65 @@
...
+
+            String keyword = input.substring(firstSpaceIndex + 1);
+
+            for (TasksDefault task : taskManager.getTasksList()) {
+                if (task.getTaskDescription().contains(keyword)) {
+                    findTasksList.add(task);
+                }
+            }
+
+            System.out.println("_____________________________________________");
```
chanjunlin: I noticed that this string is typed out multiple times across different files. Would it be beneficial to define it as a final constant in your UI file? This way you can reference wherever needed and making future updates might be easier 🤔?

src/main/java/exceptions/InvalidCommandException.java
```diff
@@ -0,0 +1,13 @@
+package exceptions;
+
+public class InvalidCommandException extends Exception {
```
chanjunlin, line 3: Maybe adding Javadoc comments to describe the class's purpose and usage? This could help people understand when this exception is thrown 🤔?

chanjunlin (commented):
LGTM. Good work on the Luigi implementation! The structure is clear and i appreciate your attention to error handling as well 😄

PR #130, opened 2025-01-27
[curiousfun88] iP

curiousfun88:
# Blob!!

> “I am opposite of Artificial Intelligence. I am Genuine Stupidity.” – Unknown [(source)](https://example.com)

Blob helps you do the things that you are too lazy to do yourself. It's,

- text-based
- easy to learn
- ~FAST~ *SUPER FAST* to use

All you need to do is,

1. download it from [here](https://github.com/curiousfun88/ip).
2. double-click it.
3. add your tasks.

And it is *FREE!*  

## Features:

- [x] Managing tasks  
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}


src/main/java/blob/Blob.java
crmlatte, line 1: Nice! I think the java docs are good and coding standard is good

src/main/java/blob/TaskList.java
crmlatte, line 1: The method and variable naming looks gd!

src/main/java/blob/Parser.java
crmlatte, line 1: Maybe can use switch statement here but is easy to read too!

src/test/java/blob/StorageTest.java
crmlatte, line 1: Method names are clear!

crmlatte (commented):
Overall your code looks easy to read and understand. gd job

PR #133, opened 2025-01-27
[LonelyFort] iP

LonelyFort:
# Angela

A straightforward and convenient task and reminder app which you can count on to keep track of the never-ending responsibilities in your life.

It's:
- [x] Ad-free
- [x] Free of charge
- [x] Non-subscription based (surprisingly hard to come by)
- [x] Clear and straightforward text-commands
- [x] Runs on Java (Because developing applications on JavaScript is for cowards) 
- [x] Open Source 
- [x] No need to install third party library  
- [x] Has nice art when you boot up the app

## Features (Ver 1.2)
- Manage Todo, Deadline and Events capable of handling dates.
- Load and store tasks onto local device.
- Filter tasks that contains specific keywords. ~~(In Progress)~~ **(Available in v1.1)**
- Graphic User Interface for users.

## Installation Instructions
1. If you have not installed Java 17 onto your local device, you may download it from [here](https://www.oracle.com/java/technologies/javase/jdk17-archive-downloads.html).
2. Download latest release of Angela.jar from [here](https://github.com/LonelyFort/ip/releases).
3. Place the jar file into an empty folder (Preferably under the same name Angela).

## Run Instructions
4. Once the file is placed, open the terminal by right click -> **Open In Terminal**.
5. Enter the following into the terminal window to run the program: `java -jar .\Angela.jar`

## Disclaimer
As a new file will be created when the program is running, please ensure you do not set the root file to require admin permission for the program to run smoothly.

## Random Quote Because the Requirement Says So (And a random Emoji too)

> "That's that and this is this. 😏 " - Some washed-up Grade 9 Fixer.


src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param filePath The path to the file that will be used by the `Database` for storage or retrieval.
+     *                 This path is passed to the `Database` constructor.
+     * @see TaskList
+     * @see Database
+     */
+    public Angela(String filePath) {
+        setTimeout(() -> {
+            this.listData = new TaskList();
+            this.database = new Database(filePath, listData);
+        },10000);
```
Reshiro:
```suggestion
        }, 10000);
```
Commas should be followed by a white space.

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+     * The input is stripped of leading and trailing whitespace before being processed.
+     * The method runs indefinitely in a loop, handling chat commands via the chatResponse method.
+     */
+    private void echo() {
+        Scanner scan = new Scanner(System.in);
+
+        while (true) {
+            String input = scan.nextLine().strip();
+            // ignores comments and empty lines
+            if (input.isEmpty() || input.charAt(0) == '/') continue;
```
Reshiro: We're required to always use braces for if statements unfortunately.

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+                I will provide practical advice and emotional support until you get used to your work.
+                
+                What is your request?                        
+                """
+        ), 7000);
+    }
+
+    //@@author Oleg Mikhailov-reused
+    //Reused from https://stackoverflow.com/questions/26311470/what-is-the-equivalent-of-javascript-settimeout-in-java
+    public static void setTimeout(Runnable runnable, int delay){
```
Reshiro:
```suggestion
    public static void setTimeout(Runnable runnable, int delay) {
```
To be honest I'm not sure about the proper procedures for editing reused code, but if it's possible then a space would be a good idea here.

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+        new Thread(() -> {
+            try {
+                Thread.sleep(delay);
+                runnable.run();
+            }
+            catch (Exception e){
```
Reshiro:
```suggestion
            catch (Exception e) {
```

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+                System.err.println(e);
+            }
+        }).start();
+    }
+    //@@author
+
+    /**
+     * Main method for Angela.
+     * @param args
```
Reshiro: Remember to add an empty line before the parameters section, but I think that the param annotation here is not needed at all.

src/main/java/Angela/UI/UI.java
```diff
@@ -0,0 +1,50 @@
+package Angela.UI;
+
+/**
+ * UI class to handle display of responses in the console.
+ */
+public class UI {
```
Reshiro: I think the proper naming is Ui? But I think it's best to look at the acronym naming guidelines because it's not that clear.

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+                        
+                        
+                Welcome, Manager. I am your assistant AI, Angela.
+                I am here to support you as it's your first day. 
+                I will provide practical advice and emotional support until you get used to your work.
+                
+                What is your request?                        
+                """
+        ), 7000);
+    }
```
AnWe11: Amazing ASCII art!

But wouldn't this be better placed in the UI class?

src/main/java/Angela/Angela.java
AnWe11: Why is there a need to check for '/'? Is it in case the user forgets the command word?

src/main/java/Angela/exceptions/taskcreation/DateOrderException.java
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Returns a string representation of the exception,
+     * indicating that the end date and time cannot be earlier than the start date and time.
+     *
+     * @return a string representation of the exception
+     */
+    @Override
+    public String toString() {
+        return "End date and time cannot be earlier than the start date and time. We are not in a time loop.";
```
AnWe11: What if we are... D:

src/main/java/Angela/Angela.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    /**
+     * Prints onto the terminal when a user first initiated Angela.
+     * Utilised setTimeout method to provide a more realistic experience.
+     * ASCII art depicts Angela. from Lobotomy Corp.
+     */
+    private void greet() {
+        setTimeout(() -> System.out.println("Initiating..."), 1000);
```
Reshiro: I like the use of delays for the welcome message.

src/main/java/Angela/storage/Database.java
```diff
@@ -0,0 +1,104 @@
+package Angela.storage;
+
+import java.io.*;
```
Reshiro: IntelliJ might have added a wildcard import here automatically.

src/main/java/Angela/tasktype/Task.java
```diff
@@ -0,0 +1,91 @@
...
+    public void uncheck() {
+        this.isCompleted = false;
+    }
+
+    /**
+     * Returns the completion status of the task.
+     *
+     * @return True if the task is completed, otherwise false.
+     */
+    public boolean isCompleted() {
```
Reshiro: Should the method name be more like a verb here?

src/main/java/Angela/storage/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+        for (int i = 0; i < size(); i++) {
+            String taskToBeSaved = get(i).toSaveFormat();
+            if (i < size() - 1) {
+                taskToBeSaved += "\n";
+            }
+            stringBuilder.append(taskToBeSaved);
+        }
+        return stringBuilder.toString();
+    }
```
AnWe11: This is an extremely elegant way to save the tasks, good job!

Reshiro (commented):
The code is very high quality, and I like the Lobotomy Corp reference. There's just a few minor deviations from the recommended coding standard.

src/main/java/Angela/UI/UI.java
```diff
@@ -0,0 +1,50 @@
+package Angela.UI;
+
+/**
+ * UI class to handle display of responses in the console.
```
AnWe11: Although your explanation for your comments are crystal clear, maybe your comments can be more succinct?

src/main/java/Angela/UI/UI.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Displays response in a simple 2 line UI.
+     *
+     * @param response The string to be displayed.
+     */
+    public static void displayResponse(String response) {
+        System.out.println("________________________________________________________\n"
+                + response + "\n"
+                + "________________________________________________________");
+    }
```
AnWe11: Is there a need for a documentation here for a simple function?

AnWe11 (commented):
I think your code is extremely organised and I have a lot to learn from you. I like the way you packaged and compartmentalised your classes and even for your exceptions. The way you read from the textfile to update the taskList was also innovative. Overall, great job on this and Angela will definitely be a hit.

PR #142, opened 2025-01-27
[gekjunxu] iP

gekjunxu:
# Bob Bot 🤖 

> **Bob** the Builder: Can We Fix it? Yes We Can! - **Bob** the Builder

Bob frees your mind of having to remember what you need to do. It is
- Text Based
- Java Based
- Bob Based
- **Freeeeeee~eee~**!! 😱 

To get started, simply:
1. Head to the Releases [page ](https://github.com/gekjunxu/ip/releases)to download the .jar file
2. Double click
3. Enjoy!

## Features 

- [x] Managing your todo list
- [x] Managing Deadlines
- [x] Keeping track of Events
- [x] Easily search your list for your wanted task

### Challenge yourself 🦾 
If you want to change the location of the data file, edit the `main` method in the Bob.java file:
```Java    
public static void main(String[] args) throws IOException, BobException {
        new Bob("data/bob.txt").run();
    }
```
Change the `data/bob.txt` part to your desired location. (Remember, use _relative_ paths!)


src/main/java/bob/Bob.java
```diff
@@ -214,7 +212,19 @@ public void run() throws IOException {
                         ui.printNumOfItemsInList(taskList.size());
                         ui.printLine();
 
-                    } else {
+                    } else if (input[0].equalsIgnoreCase("find")) {
+                        if (input.length < 2) {
+                            throw new BobException("Incorrect number of arguments");
```
johngao122: Have you considered using a switch statement instead of if else

src/main/java/bob/task/TaskList.java
```diff
@@ -48,9 +48,26 @@ public void deleteTask(int index) {
...
+     * @param name To find in description of tasks.
+     * @return TaskList containing the found tasks.
+     */
+    public TaskList findTasks(String name) {
+        ArrayList<Task> foundTasks = new ArrayList<>();
         for (Task task : tasks) {
-            System.out.println(task);
+            if (task.description.contains(name)) {
+                foundTasks.add(task);
+            }
```
johngao122, line 73: Is direct field access intended here? Perhaps we should use a getter method to maintain encapsulation?

src/main/java/Bob.java
```diff
@@ -190,4 +192,48 @@ public static void main(String[] args) throws BobException, IOException {
 
 
     }
-}
+
+
+    private static Task parseTask(String line) throws BobException {
+        char taskType = line.charAt(1); // The character after the first '['
```
johngao122: Should validate line length before accessing characters.

johngao122 (commented):
I like the improvements to code organization and readability, particularly the extraction of parsing logic into a dedicated method. However, there are some concerns around error handling and input validation that should be addressed before merging.

PR #157, opened 2025-01-27
[kumar2215] iP

kumar2215:
RoverPro frees your mind of having to remember things you need to do. It's,
+ text-based
+ easy to learn
+ ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://github.com/kumar2215/ip/releases).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 💯

And it is **FREE**!

Features:
- [X] Managing tasks
- [X] Managing deadlines
- [ ] Reminders (coming soon)

```java
public final class Rover {
        new Rover("data/Rover.txt").run();
    }
}
```


src/test/java/rover/task/EventTest.java
```diff
@@ -0,0 +1,143 @@
+package rover.task;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
darrenchooji, line 2: Any reason why you chose to 'import static' rather than 'import'?

src/test/java/rover/task/TaskListTest.java
```diff
@@ -0,0 +1,209 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.time.format.DateTimeParseException;
+
+import org.junit.jupiter.api.Test;
+
+import rover.exceptions.RoverException;
+import rover.ui.Ui;
+
+public class TaskListTest {
```
darrenchooji, line 15: I like how comprehensive your testing is.

src/main/java/rover/ui/Ui.java
```diff
@@ -0,0 +1,109 @@
...
+        System.out.println(divider);
+    }
+
+    /**
+     * Displays the help message when the user types an invalid command.
+     */
+    public void showHelpMessage() {
+        System.out.println(divider);
+        System.out.println("I'm sorry, but I don't know what that means.");
+        String briefHelp = """
```
darrenchooji: I like the additional help provided, users would find this helpful.

src/test/java/rover/task/EventTest.java
kumar2215: It doesn't work if it's not a static import?

src/test/java/rover/task/TaskListTest.java
kumar2215: Thanks!

src/main/java/rover/ui/Ui.java
kumar2215: Thanks!

PR #164, opened 2025-01-27
[laishuya] iP

laishuya:
# Luigi 🔫 
> _"Yahoo!"_  

Luigi helps you remember the tasks you have to do. It's, 
- text-based
- easy to use
- easy to download

All you need to do is,
1. download the jar file from [here](https://github.com/laishuya/ip/releases/tag/v0.1/)
2. type `java -jar luigi.jar` on your terminal to run Luigi
3. add your tasks
4. let it manage your tasks for you 🤯 

And it is **FREE**!

Features:
- [x] Add todo, deadline and event tasks
- [x] Find tasks with the same keyword or date
- [x] Ability to mark, unmark and delete tasks

> [!NOTE]
> User guide on how to navigate Luigi chatbot coming soon!

> [!WARNING]
> The main class below is still a work in progress and is not yet functional. This page will be updated when the main class is complete.

If you are a Java programmer, you can use it to practise Java. Here's the `main` method:
```ruby
public class Main {
    }
}
```


src/main/java/luigi/TaskList.java
```diff
@@ -0,0 +1,182 @@
...
+        return sb.toString();
+    }
+
+    /**
+     * Unmarks a Task as completed.
+     *
+     * @param index The position of the Task in the list.
+     * @return A string of the unmarked Task.
+     */
+    public String unmark(int index) {
```
panomia01, line 227: Function name should be written in camelCase

src/main/java/luigi/commands/EventCommand.java
```diff
@@ -0,0 +1,42 @@
...
+import luigi.Storage;
+import luigi.TaskList;
+import luigi.ui.Ui;
+
+/**
+ * Represents a command to add an Event to the TaskList.
+ */
+public class EventCommand extends Command {
+    private final String description;
+    private final String from;
```
panomia01: perhaps a noun is more intuitive

src/main/java/luigi/commands/EventCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    private final String to;
```
panomia01: same for this too!

src/main/java/luigi/ui/MainWindow.java
```diff
@@ -0,0 +1,65 @@
...
+    private Luigi luigi;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/Mario.png"));
+    private Image luigiImage = new Image(this.getClass().getResourceAsStream("/images/Luigi.png"));
+
+    @FXML
+    public void initialize() {
+        String greeting = "Hello! I'm Luigi!\nHow can I help you?";
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+        dialogContainer.getChildren().add(DialogBox.getLuigiDialog(greeting, luigiImage));
```
panomia01, line 40: Love that you have images to your chatbot

panomia01 (commented):
Overall, I found your code easy to read. Only have some naming conventions

src/main/java/luigi/Luigi.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Represents Luigi the Chatbot, which is initialised with the given filepath.
+ * Luigi helps users to organise their tasks, by processing user commands and carrying out different functionalities,
+ * such as adding, deleting, marking, unmarking and finding tasks.
+ *
+ */
+public class Luigi {
+    private static String FILE_PATH = "./data/luigi.txt";
+    private Storage storage;
+    private TaskList list;
```
HarishB99:
```suggestion
    private TaskList tasks;
```

src/main/java/luigi/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+import luigi.tasks.Event;
+import luigi.tasks.Task;
+import luigi.tasks.Todo;
+
+/**
+ * Loads and saves user data from the data file stored on the computer.
+ * If the data file and directory are missing, then they will be created by the program.
+ */
+public class Storage {
+    private static ArrayList<Task> list = new ArrayList<>();
```
HarishB99, line 20:
```suggestion
    private static ArrayList<Task> tasks = new ArrayList<>();
```

src/main/java/luigi/TaskList.java
```diff
@@ -0,0 +1,182 @@
...
+    public String findAllTasksWithSameDate(String date) {
+        StringBuilder sb = new StringBuilder();
+        String inputFormat = "yyyy-MM-dd";
+        DateTimeFormatter format = DateTimeFormatter.ofPattern(inputFormat, Locale.ENGLISH);
+        LocalDate targetDate = LocalDate.parse(date, format);
+        for (Task task : list) {
+            if (task instanceof Deadline) {
+                Deadline deadline = (Deadline) task;
+                if (deadline.getLocalDate().equals(targetDate)) {
+                    sb.append(deadline + "\n");
```
HarishB99:
```suggestion
                    sb.append(deadline + System.lineSeparator());
```

HarishB99 (commented):
In overall, I liked your code naming conventions as they were easy to read and follow. Your code also looks well organised and clean. Just a few suggestions:

1. Maybe you could use System.lineSeparator() instead of "\n"?
2. "list" might be sounding a bit too vague. I noticed in other parts of the codebase you referred to it as "tasks" or "taskList" which sounds more descriptive of what the list holds. Maybe you can try switching to the latter?

Otherwise, a job well done!

src/main/java/luigi/Luigi.java
laishuya: You're right. Thanks for pointing it out!

src/main/java/luigi/TaskList.java
laishuya: Wow, I didn't know \n is platform dependent, while System.lineSeparator() is platform independent. Thanks!

src/main/java/luigi/commands/EventCommand.java
laishuya: That makes sense, thanks for the input!

src/main/java/luigi/ui/MainWindow.java
laishuya: Ty! Good luck for the rest of your ip and tp :)

PR #174, opened 2025-01-28
[namprice227] iP

namprice227:
# Charlie
New type of **Chatbot** that:
- Help you remember task :grinning:
- Got features deadline, event, todo
- Friendly chatbot

What you need to do?
1.  Download the latest release from here [Charlie](https://github.com/namprice227/ip/releases/tag/A-Jar)
2. Run it with  
```java -jar charlie.jar```
4. Very beginner-friendly, type anything for help

It is similar to you ticking your own tickbox:
- [ ] deadline cs2103t home work /by 2025/02/06
- [x] todo housework

If you want to programmer remember to include this:
``` java
public class Main {
    }
}
```


src/main/java/charlie/Charlie.java
```diff
@@ -0,0 +1,30 @@
+package charlie;
+
+/**
+ * The main class that orchestrates the Charlie task manager.
+ * It initializes storage, loads tasks, and processes user commands.
+ */
+
+public class Charlie {
```
huangtian666, line 8: I like how you have coded in a more OOP way like separating into different components such as UI, Task and so on!

src/main/java/charlie/Charlie.java
```diff
@@ -0,0 +1,30 @@
...
+        while (true) {
+            try {
+                String fullCommand = Ui.readCommand();
+                Parser parser = new Parser(fullCommand);
+                parser.action(taskList);
+            } catch (Exception e) {
+                Ui.showError(e.getMessage());
+            }
+        }
+    }
```
huangtian666: It is nice to see that your implementation is so clean and simple!

src/main/java/charlie/Ui.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public static void printHelp() {
+        System.out.println("Here are the commands you can use:");
+        System.out.println("1. todo <task description>");
+        System.out.println("2. deadline <task description> /by <time>");
+        System.out.println("3. event <event description> /from <start time> /to <end time>");
+        System.out.println("4. delete <task number>");
+        System.out.println("5. list");
+        System.out.println("6. bye");
+    }
```
huangtian666: Your help method is really helps to improve the user experience!

src/main/java/charlie/Parser.java
```diff
@@ -0,0 +1,83 @@
...
+            }
+            case "find" -> {
+                tasklist.findTask(task.trim());
+            }
+            case "bye" -> {
+                Ui.showGoodbye();
+                exit(0);
+            }
+            default -> Ui.printHelp();
+        }
```
huangtian666, line 89: Your choice of switch statements and lambda expression have effectively simplified the code and increased readability for readers!

src/main/java/charlie/Charlie.java
```diff
@@ -0,0 +1,30 @@
+package charlie;
```
huangtian666, line 1: One suggestion is that you could split the files into more packages so all relevant components are in the same folder!

src/main/java/charlie/Todo.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class Todo extends Task {
+    public Todo(String activity) {
+        super(activity);
+    }
+
+    public Todo(String activity, Boolean marked) {
+        super(activity, marked);
+    }
+
```
huangtian666, line 11: One minor suggestion here is to include @Override when you override a method in the superclass 😊

huangtian666 (commented):
Overall, nice implementation so far!

src/main/java/charlie/CharlieHandler.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Validates a Deadline task description.
+     * Ensures the description contains a deadline keyword (`/by`)
+     * and that the date is in the correct format (YYYY-MM-DD).
+     *
+     * @param words The Deadline task description.
+     * @return {@code true} if the description is valid, {@code false} otherwise.
+     */
+
```
TheEnd-alr-taken, line 51: I feel that the blank line between the documentation block and the method could be omitted to standardise with other areas of the code.

src/main/java/charlie/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+            case "find" -> {
+                taskList.findTask(task.trim());
+            }
+            case "bye" -> {
+                Ui.showGoodbye();
+                exit(0);
+            }
+            default -> Ui.printHelp();
+        }
```
TheEnd-alr-taken: As mentioned by @huangtian666 , I also like the choice of using lambda expression here. Any reason why you choose to separate this action(TaskList taskList) method from the getResponse(TaskList taskList) method? Both seems to have the same commands being run except for the returning of the responses.

src/main/java/charlie/Charlie.java
TheEnd-alr-taken: To add on to what @huangtian666 mentioned, I think it will look neater if you could create a charlie.task package and group up all the relevant components together since you have already taken the time to extract task specific classes.

TheEnd-alr-taken (commented):
Overall, looks good to me. Congrats on making this much progress.

src/main/java/charlie/Charlie.java
namprice227: Thank you =)

src/main/java/charlie/Charlie.java
namprice227: Thank you for your suggestion, I will consider this

src/main/java/charlie/Parser.java
namprice227: Oh, I was implementing the GUI, so I need to return the string instead of printing it out. I think I will delete the action method in the future as they said can delete the cli

PR #175, opened 2025-01-28
[Joshua Gan Yi En] iP

dragonesejosh:
Current Features:

1. Unoriginal personality
2. ~Some~ Too much OOP
3. _Basic_ task management

Other (Upcoming) Features:

- [ ] May or may not add personality
- [ ] Bare-bones GUI

Why download it?

- To easily manage your tasks ❌
- Free from bugs ❌
- Headaches from horrible UI ✅
- Most bland ChatBot known to man ✅

To keep your sanity, add this line of code to the `main` method:

``` java
    // ADD THIS LINE
    return;
    // Make sure this code never sees the light of day
    new Yale(TASKS_FILE).run();
}
```

To make a far better ChatBot yourself, here's the [original](https://github.com/nus-cs2103-AY2425S2/ip).


src/main/java/yale/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+            return;
+        }
+        for (Command cmd : COMMANDS) {
+            if (cmd.tryCommand(taskList, storage, msg)) {
+                storage.writeTasks(taskList.getTasks());
+                return;
+            }
+        }
+        tryFindCommand(msg);
+    }
```
Meatsushi64: this is an interesting implementation. why did you choose to try the command through all the messages, instead of using a case switch using an enum of COMMANDS?

src/main/java/yale/Storage.java
```diff
@@ -0,0 +1,52 @@
...
+    public ArrayList<Task> readTasks() {
+        ArrayList<Task> tasks = new ArrayList<>();
+        try (Scanner in = new Scanner(file)) {
+            while (in.hasNextLine()) {
+                Task task = Task.fromCsv(in.nextLine());
+                if (task != null) {
+                    tasks.add(task);
+                }
+            }
+        } catch (Exception ignored) {}
```
Meatsushi64:
```suggestion
        } catch (Exception ignored) {
            throw new YaleException(ignored.getMessage);
        }
```

i think it would be good to throw an exception to surface up to the Yale bot so that the user can see that there is a problem reading the file if any

but between you and me, yolo

src/main/java/yale/Task.java
```diff
@@ -0,0 +1,130 @@
...
+        return name;
+    }
+
+    public void setDone(boolean done) {
+        this.isDone = done;
+    }
+
+    public String toString() {
+        return "[%s] %s".formatted(isDone ? "X" : " ", name);
+    }
```
Meatsushi64, line 28: massive big brain to put the isDone within the toString() instead of making a whole new method 🆙 🔝

Meatsushi64 (commented):
Overall love the memes in your PR description. 

I have no idea what it is with the UI stuff but I guess I'll find out soon...

Maybe would be good to surface any exception if any to the user

cheers man 👍

src/main/java/yale/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+            new Command("deadline", "deadline [name] /by [date]",
+                    "deadline (.+) /by (.+)",
+                    (t, m) -> t.addTask(new Task.Deadline(m.group(1), m.group(2)))),
+            new Command("event", "event [name] /from [start] /to [end]",
+                    "event (.+) /from (.+) /to (.+)",
+                    (t, m) -> t.addTask(new Task.Event(m.group(1), m.group(2), m.group(3)))),
+            new Command("delete", "delete [id]", "delete (\\d+)",
+                    (t, m) -> t.deleteTask(Integer.parseInt(m.group(1)))),
+            new Command("find", "find [keyword]", "find (.+)",
+                    (t, m) -> t.listSearch(m.group(1)))
```
ernestlsy: Should these lines be indented by 4 spaces instead?

src/main/java/yale/Storage.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Reads in the tasks in the task file and outputs them.
+     *
+     * @return ArrayList of the tasks in the file.
+     */
+    public ArrayList<Task> readTasks() {
+        ArrayList<Task> tasks = new ArrayList<>();
+        try (Scanner in = new Scanner(file)) {
```
ernestlsy, line 32: Nice use of try-with-resources statement to handle closing of the Scanner!

src/main/java/yale/Ui.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class Ui {
+    private static final String NAME = "Yale";
+
+    private StringBuilder output;
+
+    /**
+     * Creates a Ui to display output.
+     */
+    public Ui() {}
```
ernestlsy: Declaring this default constructor seems unnecessary.

ernestlsy (commented):
LGTM! No significant issues besides a few nits

PR #176, opened 2025-01-28
[basil-boh] iP

basil-boh:
### Steve 
 

> ~ _Do-do-do do do-do do, your friend....Steve_! - Rent-free Tiktok Sound Track 🔥

**Steve is your all in one friend that keeps track of tasks for you! 🫵🏻** 
- Intuitive commands ✅
- Available 24/7 🕔
- Lighting Fast ⚡

**To get bad boy Steve going, you need to:**

1. Download it from here
2. Double click it
3. Use by entering commands followed by task description
4.  Leave the rest to Steve ❤️

So...what are you waiting for? Try Steve out for free!!! 🤓

Features:

- [ ] 1. To do tasks 
- [ ] 2. Deadline tasks
- [ ] 3. Event Tasks
- [ ] 4. List
- [ ] 5. Mark/Unmark
- [ ] 6. Delete

```
package steve.ui;

import steve.tasks.Task;
import steve.tasks.TaskManager;
import steve.ui.UserInterface;
import steve.storage.FileHandler;
import java.util.ArrayList;

// Main Class File
public class Steve{
        FileHandler fileHandler = new FileHandler("src/main/data/steve.txt");
        ArrayList<Task> tasks = fileHandler.loadTasks(); // Load tasks from file
        TaskManager taskManager = new TaskManager(tasks); // Initialize TaskManager with loaded tasks
        UserInterface ui = new UserInterface(taskManager); // Initialize UserInterface with TaskManager

        // Start the application
        ui.start();
    }
}
```


src/main/java/steve.tasks/Task.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.ArrayList;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+
+// General Task
+public class Task {
+    protected String description;
+    protected boolean done;
+    protected static int taskCount = 0;
```
JohannsenLum, line 14: I'm not sure if your static taskCount would be still needed as you have your TaskManager to handle the list of tasks

JohannsenLum (commented):
Your code is well-structured and follows good practices!

build.gradle
```diff
@@ -0,0 +1,44 @@
...
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.10.0'
+    testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.10.0'
+    implementation group: 'com.joestelmach', name: 'natty', version: '0.13'
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.10.0'
+    testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.10.0'
```
Jovin-Ang: Opps :sweat_smile:, I think this looks like duplicate lines

src/main/java/steve.commands/Command.java
```diff
@@ -0,0 +1,7 @@
+package steve.commands;
+import steve.exceptions.InvalidTaskNumberException;
```
Jovin-Ang, line 2: Perhaps there should be an empty line in between here. I noticed this in `ListCommand.java` too.

src/main/java/steve.commands/CommandCreate.java
```diff
@@ -0,0 +1,32 @@
...
+                return new TodoCommand(taskManager, userInput);
+            case "deadline":
+                return new DeadlineCommand(taskManager, userInput);
+            case "event":
+                return new EventCommand(taskManager, userInput);
+            case "delete":
+                return new DeleteCommand(taskManager, userInput);
+            default:
+                throw new InvalidCommandException("Unknown command: " + firstWord);
+        }
```
Jovin-Ang: I believe that the case clauses should have **no** indentation according to the coding standard

src/main/java/steve.commands/MarkCommand.java
```diff
@@ -0,0 +1,23 @@
...
+        this.userInput = userInput;
+    }
+
+    @Override
+    public void execute() throws InvalidTaskNumberException, NumberFormatException {
+        int taskNumber = Integer.parseInt(userInput.split(" ")[1]);
+        taskManager.markTask(taskNumber, true); // Mark as done
+    }
+
+
```
Jovin-Ang: Are these extra line breaks intended?

src/main/java/steve.tasks/Deadline.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+            } else {
+                this.description = "Invalid format. Usage: deadline <description /by yyyy-MM-dd HHmm>";
+                super.decreaseTaskCount();
+            }
+        } else {
+            this.description = "Description cannot be empty. Usage: deadline <description /by yyyy-MM-dd HHmm>";
+            super.decreaseTaskCount();
+        }
+    }
```
Jovin-Ang: This part is a little unclear to me. Is this error message supposed to be displayed to the user or stored as the task description? Perhaps you could add some code comments?
Since this is a constructor, the Deadline object will still be created provided an exception is thrown. Perhaps you could extract the checking logic elsewhere and only pass valid parameters to the constructor.
Similar comment for the other types of tasks.

src/main/java/steve.tasks/Task.java
```diff
@@ -0,0 +1,118 @@
...
+                    "Task number " + taskNumber + " is invalid. Please enter a number within Task Count Range: " + taskCount
+            );
+        }
+        else {
+            Task task = userData.get(taskNumber - 1);
+            task.markDoneOrNot(command == CommandType.Mark);
+            task.message(firstWord);
+        }
+    }
+    public static void validateTaskNumberDel(int taskNumber, int taskCount, ArrayList<Task> userData) throws InvalidTaskNumberException {
```
Jovin-Ang: Missing empty line in between

src/main/java/steve/ui/Messages.java
```diff
@@ -0,0 +1,35 @@
...
+        System.out.println("____________________________________________________________\n" +
+                "     \n" + "Bye! Hope to see you again soon =)" + "\n" +
+                "____________________________________________________________");
+    }
+
+    public static void unknown() {
+        String first = "Sorry I don't quite understand!";
+        String second = " Please add a task by entering: \n 1. todo \n 2. deadline \n 3. event \n 4. find \n";
+        String third = "Followed by a description of your task! =)";
+        System.out.println(first + second + third);
```
Jovin-Ang: Perhaps you can condense this into a single multi line string. Likewise for several other cases.
```suggestion
        System.out.println("""
        Sorry I don't quite understand! Please add a task by entering: 
        1. todo
        2. deadline
        3. event
        4. find
        Followed by a description of your task! =)""");
```

src/main/java/steve.tasks/Deadline.java
```diff
@@ -0,0 +1,87 @@
+package steve.tasks;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+// Deadline task
+public class Deadline extends Task {
+    protected String description;
```
Jovin-Ang: There is no need to redeclare a instance variable that is already present in the superclass.

Jovin-Ang (commented):
I noticed the task classes are a little long. Perhaps it can be further simplified by removing some unnecessary methods which are not used and placing common instance variables into the `Task.java` superclass, such as `isValid`, which seems to be present in all 3 subclasses.

build.gradle
GrassHeadd: As mentioned by Jovin, I think there are duplicated lines, I assume you probably copied the tutorial's 2 lines below much as I did, I think maybe can review more carefully next time to see that its mostly already in the original repo!

src/main/data/steve.txt
GrassHeadd, line 1: actually this file could be just added to gitignore since given an actual todo chatbot, you would probably not want to have your own txt files be kept in others' computer, would help to just add this line
```
steve.txt
```
to tell github to ignore the changes you made locally to this file

src/main/java/steve.commands/Command.java
GrassHeadd, line 2: as mentioned by jovin, from my vague memory of 2030s and just general java programming, the general consensus for this scenario is to separate the package line and imports with an empty line in between

src/main/java/steve.commands/Command.java
GrassHeadd, line 1: just wanted to add that the commands file is really well abstracted!

src/main/java/steve.commands/FindCommand.java
GrassHeadd, line 1: just curious about why you didnt add a description to this class, also i believe we are suppose to add javadocs to the whole class too

src/main/java/steve.tasks/Deadline.java
```diff
@@ -0,0 +1,87 @@
...
+            result = this.description + " (By: " + formatDateTime(this.by) + ") ";
+        } else {
+            result = this.description;
+        }
+        return result;
+    }
+
+    @Override
+    public String getOriginalDescription() {
+        return this.description + " /by " + (this.by != null ? this.by.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm")) : "");
```
GrassHeadd: can split this into 2 lines since its kind of long

src/main/java/steve.tasks/Task.java
```diff
@@ -0,0 +1,118 @@
...
+                : "____________________________________________________________\n" +
+                "     OK, I've marked this task as not done yet:\n" +
+                "       [ ] " + this.description + "\n" +
+                "____________________________________________________________";
+        System.out.println(result);
+    }
+
+    public void markDoneOrNot(boolean isDone) {
+        this.done = isDone;
+    }
```
GrassHeadd: name feels unconventional, maybe could have put toggleDone or smth

src/main/java/steve.tasks/Task.java
```diff
@@ -0,0 +1,118 @@
...
+    public int taskCount() {
+        return taskCount;
+    }
+    public void decreaseTaskCount() {
+        taskCount--;
+    }
+
+    public String list() {
+        return "." + " [" + this.code() + "]" + "[" + this.getStatusIcon() + "] " + this.getDescription();
+    }
```
GrassHeadd, line 136: i feel listTask is a better name but list is fine too!

src/main/java/steve.tasks/ToDo.java
```diff
@@ -0,0 +1,65 @@
...
+
+// Todo test
+public class ToDo extends Task {
+    protected String description;
+    protected boolean isValid = false;
+
+    public ToDo(String description) {
+        super(description == null || description.trim().isEmpty()
+                ? "Description cannot be empty. Usage: todo <description>"
+                : description);
```
GrassHeadd: can maybe stop them from even initialising the tasks by using a try catch block rather than letting the task classes handle it

src/main/java/steve/storage/FileHandler.java
```diff
@@ -0,0 +1,51 @@
...
+public class FileHandler {
+    private String filePath;
+
+    public FileHandler(String filePath) {
+        this.filePath = filePath;
+    }
+
+    private static Task parseTask(String line) {
+        String firstWord = line.split(":")[0];
+        String restOfLine = line.substring(firstWord.length() + 1).trim();
```
GrassHeadd, line 41: continuation of the above comment, maybe can put smth like but honestly i think your implementation is fine, just an alternative i guess that i personally would do
```
try {
  String restOfLine = line.substring(firstWord.length() + 1).trim();
} catch (NullPointerException e) {
  System.out.println("Description cannot be empty.")
}
```

GrassHeadd (commented):
Overall really well done! Note worthy things are the good abstractions of the different classes and packages, also created your own exceptions, however does not seem to add javadocs to the files and could have utilised the built in exceptions better. Great work!

src/main/data/steve.txt
basil-boh: rly good idea I completely forgot about gitignore haha

src/main/java/steve.commands/Command.java
basil-boh: thank you :)

src/main/java/steve.commands/Command.java
basil-boh: thanks for pointing that out!

src/main/java/steve.commands/CommandCreate.java
basil-boh: You're right will take note of this

src/main/java/steve.commands/FindCommand.java
basil-boh: truthfully I went for a break and forgot to add it afterwards hahaha but will def add it in soon

src/main/java/steve/ui/Messages.java
basil-boh: Great suggestion will adopt it

build.gradle
basil-boh: Thanks for pointing that out!

src/main/java/steve.commands/MarkCommand.java
basil-boh: Good catch thanks!

src/main/java/steve.tasks/Deadline.java
basil-boh: Agreed, will do

PR #179, opened 2025-01-28
[m0destly] iP

m0destly:
Darwin frees your mind of having to remember things you need to do. It's:

- text-based
- easy to learn
- ~~FAST~~ *SUPER* FAST to use

All you need to do is:
1. Download it from [here](https://github.com/m0destly/ip)
2. Double click it
3. Add your tasks
4. Let it manage your tasks for you 🤩 

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon) 
##
```java
public class Main {
    }
}
```


src/main/java/darwin/Parser.java
```diff
@@ -0,0 +1,69 @@
...
+            String keyword = command.substring(4).trim();
+            if (keyword.isEmpty()) {
+                throw new DarwinException(ErrorMessage.MISSING_KEYWORD.message());
+            }
+            taskList.find(keyword);
+        } else {
+            String[] inputs = command.split(" /");
+            taskList.add(inputs);
+        }
+        return false;
```
Shamanbenny: I feel like it might be easier to grasp the entire structure of this if it was set in `switch case` instead?

Shamanbenny: If I am also not mistaken, I believe `switch case` on `String` variable is done with Hash lookup, which results in an O(1) lookup rather than chaining `if else` statements which could result in O(n), where `n` is the number of conditions.
Read more about it [here](https://medium.com/@thedevelopernote/efficiency-of-if-vs-switch-in-java-a-quick-comparison-32f4ee0e20a3).

src/main/java/darwin/Storage.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
+
+public class Storage {
+    private String filePath;
```
Shamanbenny: Would it be better to use `final` to prevent accidental assignment?

src/main/java/darwin/Storage.java
```diff
@@ -0,0 +1,78 @@
...
+import task.Task;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.util.ArrayList;
```
Shamanbenny: Should these imports be ordered in accordance with the module's [Coding Standard](https://se-education.org/guides/tutorials/intellijCodeStyle.html#tweak-import-order)?

src/main/java/darwin/TaskList.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Adds a new task to the tasklist.
+     *
+     * @param inputs The string array that splits the original string into respective fields.
+     * @throws DarwinException If the format of the command is violated or command is not understood.
+     */
+    public void add(String[] inputs) throws DarwinException {
+        if (inputs[0].equals("todo") || inputs[0].startsWith("todo ")) {
```
Shamanbenny: Would it be better to extract these `if else` statements and instead, call them by their own functions instead? (E.g. `addTodo()`, etc.)

src/main/java/darwin/Ui.java
```diff
@@ -0,0 +1,32 @@
+package darwin;
+
+import java.util.Scanner;
+
+public class Ui {
+    private static final String divider = "____________________________________________________________";
```
Shamanbenny: Should this be in `SCREAMING_SNAKE_CASE` instead? Refer to [this](https://se-education.org/guides/conventions/java/intermediate.html#naming).

src/main/java/exception/ErrorMessage.java
Shamanbenny, line 1: I like your intuitive idea to use enum for error messages!

Shamanbenny (commented):
Wonderful work thus far, aside from a few nits. Keep up the effort!

src/main/java/darwin/Storage.java
```diff
@@ -25,7 +25,7 @@ public ArrayList<Task> load() throws DarwinException {
             ObjectInputStream ois = new ObjectInputStream(fis);
             ArrayList<Task> tasks = (ArrayList<Task>) ois.readObject();
             ois.close();
-            System.out.println(tasks.size() + " tasks loaded.");
+            System.out.println(tasks.size() + (tasks.size() > 1 ? " tasks loaded." : " task loaded."));
```
Fieash: I like the concise print statement, but should this be a one liner?

src/main/java/darwin/Parser.java
m0destly: Thanks for the insightful suggestion 😄

src/main/java/darwin/Ui.java
m0destly: Good catch 👍

src/main/java/darwin/Storage.java
```diff
@@ -13,7 +13,7 @@
 import java.util.ArrayList;
 
 public class Storage {
-    private String filePath;
+    private final String filePath;
```
Shamanbenny, line 20: Just glimpsed over your commits, as I was notified via email notification, I believe this should also be `FILE_PATH` as it's a constant variable.

Fieash (commented):
LGTM, some splitting of lines requested

PR #183, opened 2025-01-28
[soonami69] iP

soonami69:
# Clonky
> "I think this is very good" - Goku ([Source](Imadeitup.com))

***Clonky*** helps you store tasks and keep track of what to do.

***Clonky*** is:
- 100% Safe
- Fast
- Works (sometimes, most of the time)

All you need to do is:
1. Download it [here](https://github.com/soonami69/ip/releases/tag/A-Jar)
2. Double-click it
3. Figure the rest out yourself 🦄

Features:
- [x] Manage tasks
- [x] Search for tasks by descriptions
- [ ] Reminders
- [ ] ~~Mine bitcoin~~ Play ASMR

If you are a `Java` programmer, you can run it too<sup>*</sup>. Here's the `main` method:
```java
public class Main {
        Application.launch(Clonky.class, args);
    }
}
```
<sub>*I didn't actually test this</sub>


src/main/java/clonky/tasks/Event.java
```diff
@@ -3,16 +3,33 @@
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 
+/**
+ * The {@code Event} class represents a task that occurs within a specific time frame.
+ * It extends the {@code Task} class and stores {@code LocalDate} values for the start and end dates.
+ */
```
subiloble, line 9: I like your JavaDoc comments which is clear and specific.

src/main/java/clonky/tasks/Parser.java
```diff
@@ -23,6 +34,13 @@ public void addTodo(String description) throws NoDescriptionException {
         printTaskCount();
     }
 
+    /**
+     * Adds a new {@code Deadline} task with a due date.
+     *
+     * @param arguments The task description and due date in the format "description /by YYYY-MM-DD".
+     * @throws NoDescriptionException If the description is missing.
+     * @throws NoByException If the due date is missing.
+     */
```
subiloble, line 39: LGTM.

src/main/java/clonky/tasks/Event.java
```diff
@@ -0,0 +1,39 @@
...
+    private final LocalDate from;
+    private final LocalDate to;
+
+    /**
+     * Constructs a new {@code Event} task with a description, start date, and end date.
+     *
+     * @param description The description of the event.
+     * @param from The start date of the event in {@code YYYY-MM-DD} format.
+     * @param to The end date of the event in {@code YYYY-MM-DD} format.
+     */
```
UltimateHG, line 20: I like the style of documentation, clear and concise

src/main/java/clonky/tasks/TaskList.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param filePath The file path to save to.
+     * @return A response detailing if the operation was successful or not.
+     */
+    public Response saveTasks(String filePath) {
+        String text = "";
+        Mood mood;
+        Color color;
+        String path = (filePath == null || filePath.isEmpty())
+                ? Paths.get("clonky", "tasks.txt").toString()
+                : filePath;
```
UltimateHG: Would it be better to set a user-specifiable save path (i.e. in a config file) instead of accepting an arbitrary filepath as a fully user-controllable variable?

src/main/java/clonky/tasks/TaskList.java
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Finds tasks in the list that match the given description.
+     *
+     * @param description The description to search for.
+     * @return A list of tasks that match the given description.
+     */
+    public List<Task> findTasksByDescription(String description) {
+        return tasks.stream()
+                .filter(task -> task.description.toLowerCase().contains(description.toLowerCase()))
+                .collect(Collectors.toList());
```
UltimateHG: I like the use of functioning programming here to handle the finding of tasks 👍

UltimateHG (commented):
Clear and well-documented code, looks good.

src/main/java/clonky/tasks/TaskList.java
soonami69: Thanks for the suggestion! That's a great idea actually

PR #184, opened 2025-01-28
[zacklow28] iP

zacklow28:
### FeedMe

- easy to use like `code`
    - cause it is 
    - only difference is you get to specify a file first
        - if file is found, load that file
        - if file not found, create a new file with specified name 

1. download it [here](https://github.com/zacklow28/ip)
2. use it ✅ 

- [x] Done
- [ ] Not Done 

This is the code for $\frac {greet} {meet}$
```Java
public class Main {
        ui.greet();    
    }
}
```
> FeedMe more!

*Free* $\cup$ **Free** $\cap$ ~Free~


src/main/java/feedme/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+public class Parser {
+
+    /**
+     * Parses user input into a command
+     * @param in user input
+     * @return String command
+     */
+    public static String parse(String in) {
+        String[] inArr = in.split(" ", 2); //split just on the first space
+        if (in.equals("bye")) {
```
feliciahmq, line 17: Perhaps can use switch case instead of if else here. We can use ENUMS too.

src/main/java/feedme/Storage.java
```diff
@@ -0,0 +1,136 @@
...
+            switch (taskType) {
+            case 'T':
+                if (line.contains("[X]")) {
+                    tasklist.addTask(new ToDo(food, true));
+                } else {
+                    tasklist.addTask(new ToDo(food));
+                }
+                break;
+            case 'D':
+                String[] foodArr = food.split(" by: ");
```
feliciahmq, line 66: Maybe we can parse the line and create a new ToDo/ Deadline/ Event inside their own subclass instead of here?

src/main/java/feedme/Ui.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Processes the user input to update the tasklist
+     * @param in user input to be passed to the parser
+     * @param tasklist tasklist to be updated
+     * @param storage storage for reading from and writing to file
+     * @param br buffer reader for reading user input
+     * @throws IOException if input cannot be read
+     */
+    public void respondToUserBasedOnCommand(String in, Tasklist tasklist, Storage storage, BufferedReader br)
```
feliciahmq: Why not create specific response functions for each kind of command and call it in their command subclass? eg. in todo, call showToDo()

src/main/java/feedme/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+            return "goodbye";
+        } else if (in.equals("list")) {
```
feliciahmq, line 19: Perhaps can extract out Command general class, and have specific subclasses?

src/test/java/feedme/task/TasklistTest.java
```diff
@@ -0,0 +1,41 @@
+package feedme.task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TasklistTest {
```
feliciahmq: I like how thorough the tests are.

feliciahmq (commented):
LGTM 👍🏻

src/main/java/feedme/MainWindow.java
```diff
@@ -0,0 +1,124 @@
...
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/user.png"));
+    private Image botImage = new Image(this.getClass().getResourceAsStream("/images/charmander.png"));
+    private boolean isInitialized = false;
+
+    /**
+     * Sets up the GUI with a start message
+     */
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
```
amir-dha, line 43: Perhaps, you could introduce line break in these types of statements to enhance readability.

src/main/java/feedme/Main.java
```diff
@@ -0,0 +1,30 @@
+package feedme;
```
amir-dha, line 1: Perhaps, you could break it into more specific packaging.

src/test/java/feedme/ParserTest.java
```diff
@@ -0,0 +1,39 @@
...
+    @Test
+    public void parseInputToCommandInt() {
+        //case 1
+        assertEquals(0, Parser.parseInputToIndex("mark 1"));
+        assertEquals(1, Parser.parseInputToIndex("delete 2"));
+        assertEquals(32, Parser.parseInputToIndex("unmark 33"));
+    }
+
+    @Test
+    public void parseInputToCommandTask() {
```
amir-dha, line 31: Perhaps variable names could be more succinct? I noted this in other places too.

amir-dha (commented):
This looks really good. The style and the code are comprehensive and very readable.

PR #206, opened 2025-01-29
[Ming Sheng] iP

moingshoing:
### Aris

> "I don't know my purpose here." - Jae Crowder on his role with the Bucks

Aris frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://github.com/moingshoing/ip/releases).
2. double-click it.
3. add your tasks.

Please send me money I'm **broke**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

`"Hello world"`


src/main/java/aris/Aris.java
```diff
@@ -0,0 +1,125 @@
...
+    public Aris(String filePath) {
+        this.list = new TaskList();
+        this.storage = new Storage(filePath);
+        this.arisUi = new Ui();
+        try {
+            storage.loadFile(list);
+        } catch(FileNotFoundException e) {
+            arisUi.format("No file found ¯\\_(._.)_/¯");
+        } catch(IllegalArgumentException e) {
+            arisUi.format("File corrupted ¯\\_(._.)_/¯");
```
wallacepck: Should the program continue to run after knowing its savedata is corrupted?

src/main/java/aris/list/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+     * @return A confirmation message.
+     */
+    public String markTaskUndone(int index) {
+        try {
+            isIndexValid(index);
+            return list.get(index - 1).markUndone();
+        } catch (IllegalArgumentException e) {
+            return "Number is out of range ¯\\_(._.)_/¯";
+        }
+    }
```
wallacepck: These two functions are almost the same, would it better to combine them into one?

src/main/java/aris/ui/Ui.java
```diff
@@ -0,0 +1,23 @@
+package aris.ui;
+
+public class Ui {
+    private String hline = "____________________________________________________________\n";
+    private String lastMessage = "";
+
+    public void format(String s) {
```
wallacepck:
```suggestion
    public void printFormattedLine(String s) {
```
Perhaps a name that better describes its behaviour?

src/main/java/aris/list/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+     * @return A confirmation message.
+     */
+    public String deleteTask(int index) {
+        try {
+            Task task = list.get(index - 1); // a little roundabout, might fix in the future
+            return task.delTask(list, index);
+        } catch (IllegalArgumentException e) {
+            return "Number is out of range ¯\\_(._.)_/¯";
+        }
+    }
```
wallacepck: I believe combining these same results into an exception or a constant would improve maintainability

src/main/java/aris/list/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Adds a new task to the list.
+     * @param task The task to add.
+     * @return A confirmation message.
+     */
+    public String addTask(Task task) {
+        return task.addTask(list);
+    }
+
+    public String findTask(String keyword) {
```
wallacepck, line 132: The parameter is named `keyword` but in use there may be more than one grammatical word, perhaps a different name like `keyphrase` that alludes to this would be better?

src/main/java/aris/task/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a task with a deadline.
+ */
+public class Deadline extends Task {
+    protected String deadline;
+    protected String deadlineString;
```
wallacepck, line 12: I can not tell what the difference between `deadline` and `deadlineString` is at a glance, perhaps a clearer name for both would help?

src/main/java/aris/task/Event.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents an event task with a start and end period.
+ */
+public class Event extends Task {
+    protected String from;
+    protected String to;
+    protected String periodString;
```
wallacepck, line 13: Instead of using the ambiguous suffix "String", perhaps naming this `formattedPeriod` or a similarly descriptive name would be clearer?

src/main/java/aris/task/Task.java
```diff
@@ -0,0 +1,74 @@
...
+     * Deletes this task from the provided list.
+     * @param list The list of tasks.
+     * @param index The index of the task to delete.
+     * @return A confirmation message.
+     */
+    public String delTask(ArrayList<Task> list, int index) {
+        // del task from duke.txt
+        list.remove(index - 1);
+        return String.format("Okei this task is deleted:\n%s\nYou now have %d task(s) in the list", status(), list.size());
+    }
```
wallacepck: This violates encapsulation, is there a reason for doing this?

src/main/java/aris/task/Task.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public String status() {
+        return isDone ? ("[X] " + description) : ("[ ] " + description);
+    }
+
+    /**
+     * Formats the task for file storage.
+     * @return A formatted string representing the task for storage.
+     */
+    public String fileFormat() {
```
wallacepck, line 50: I don't like how the term "file format" is used here, perhaps it would be better to name this `getSerialisedFormat()`?

src/main/java/aris/task/Task.java
```diff
@@ -0,0 +1,74 @@
+package aris.task;
+
+import java.util.ArrayList;
+
+/**
+ * Represents a general task with basic functionalities.
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
wallacepck, line 10: Is there a reason for `Task` not being abstract?

Additionally, the fields `isDone` and `description` are constructed in subclasses instead of the smallest scope here, would it be better to delegate it to `Task` instead?

src/test/java/aris/ArisTest.java
wallacepck, line 1: Perhaps using the given test method format `featureUnderTest_testScenario_expectedBehavior()` would be more readable?

data/Aris.txt
wallacepck, line 1: Perhaps add the data folder to gitignore? Especially for release builds, it would be confusing for end-users to see data already in the program unless it is explicitly labelled as an example.

src/main/java/aris/list/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+        return s.toString();
+    }
+
+    /**
+     * Converts the list of tasks into a file-friendly format.
+     * @return A formatted string representing the tasks for storage.
+     */
+    public String toFile() {
```
wallacepck, line 43: Perhaps this should be named `toSerialisedFormat` as this function does not actually involve any files.

src/main/java/aris/list/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return s.toString();
+    }
+
+    /**
+     * Checks if the given index is valid for accessing the task list.
+     * @param index The index to check.
+     * @throws IllegalArgumentException If the index is out of bounds.
+     */
+    public void isIndexValid(int index) throws IllegalArgumentException {
```
wallacepck, line 63: Should this return a boolean instead? I would be quite saddened if a function named `is...` immediately throws an exception.

src/main/java/aris/storage/Storage.java
```diff
@@ -0,0 +1,98 @@
...
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
+     * Loads tasks from a file into the provided TaskList.
+     * @param list The TaskList to load tasks into.
+     * @throws FileNotFoundException If the file does not exist.
+     */
+    public void loadFile(TaskList list) throws FileNotFoundException, IllegalArgumentException {
```
wallacepck, line 34: What if an `IOException` is thrown when reading the file? I would prefer this to be documented or caught by the code itself.

src/main/java/aris/storage/Storage.java
```diff
@@ -0,0 +1,98 @@
...
+            Task task;
+            String input = s.nextLine();
+            String[] parts = input.split(" \\| ", 3); // split task type, isDone and task
+
+            if (parts.length != 3) {
+                throw new IllegalArgumentException();
+            }
+
+            String taskType = parts[0];
+            int isDone = Integer.parseInt(parts[1]);
```
wallacepck, line 47: Would it be better for this to be further converted to a boolean? That would prevent any potential bugs with unexpected values and make the code easier to analyse.

src/test/java/aris/ArisTest.java
```diff
@@ -0,0 +1,72 @@
...
+                + "3.[E][ ] Test Event (from: Oct 12 2025 to: Oct 13 2025)";
+        String actual = list.printList();
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testIndexOutOfRange() {
+        String actual = list.markTaskDone(4);
+        String expected = "Number is out of range ¯\\_(._.)_/¯";
+    }
```
wallacepck: Is the `assertEquals` missing here?

src/main/java/aris/task/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+        this.deadline = dateFormatter(deadline);
+        this.deadlineString = String.format(" (by: %s)", this.deadline);
+        this.isDone = (doneInt != 0);
+    }
+
+    /**
+     * Constructs a Deadline task from a description string.
+     * @param description The task description with deadline.
+     */
+    public Deadline(String description) {
```
wallacepck: I would not use the same name `description` for this parameter, as that implies the given argument is the same description in the `Task`.

wallacepck (commented):
Overall, the code is simple and straightforward, though there are some methods/variables whose name does not clearly describe their purpose in context. I have also noticed a few potential issues in the code that may affect maintainability. 

Likewise, I would suggest moving the argument parsing out of `Task` subclasses and into `Parser` to clearly separate their responsibilities.

src/main/java/aris/Aris.java
xinnnyeee, line 1: well documented methods!

src/main/java/aris/task/Deadline.java
xinnnyeee: I agree! or perhaps extract it to a method toString(), which is what a lot of people do conventionally. perhaps it helps w people's understanding! ^_^

src/main/java/aris/task/Deadline.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param fullDescription The task description with deadline.
+     */
+    public Deadline(String fullDescription) {
+        String[] details = fullDescription.split(" /by ", 2);
+        this.description = details[0];
+        this.deadline = dateFormatter(details[1]);
+        this.deadlineString = String.format(" (by: %s)", this.deadline);
+        this.isDone = false;
+    }
+
```
xinnnyeee, line 38: interesting way to handle user input directly! perhaps would be more obvious if you name the fullDescription as userInput instead? but up to you ^_^

src/main/java/aris/task/Task.java
xinnnyeee: i agree that method should be phrased as actions! maybe getListFormat()

src/main/java/aris/ui/Ui.java
```diff
@@ -0,0 +1,30 @@
...
+        System.out.println(hline + s + "\n");
+    }
+
+    /**
+     * Displays a greeting message to the user.
+     */
+    public void greet() {
+        format("Hello! I'm Aris!\n|･ω･｀) < hi.");
+    }
+
```
xinnnyeee, line 23: i like how you used these cute text-emojis!

src/test/java/aris/ArisTest.java
xinnnyeee: that's true, I should work on changing my own test method names too HAHHA

src/main/java/aris/Aris.java
```diff
@@ -0,0 +1,125 @@
...
+
+            case BYE: // Exit program
+                arisUi.exit();
+                return;
+
+            case UNKNOWN:
+                // Fallthrough
+            default: // Any other text
+                arisUi.format("Sorry forgot to code this bit ¯\\_(._.)_/¯");
+            }
```
xinnnyeee: is it better to let user know if their problem is incomplete command or wrong command?

xinnnyeee (commented):
Generally, LGTM, a very cute program which i like! you shd look at my program too HHAAHAH.

PR #209, opened 2025-01-29
[wei-song1] iP

wei-song1:
# Codsworth
> The best companion for mankind 😉
```
________________________________________________________
|   _____          _                        _   _      |
|  / ____|        | |                      | | | |     |
| | |     ___   __| |_____      _____  _ __| |_| |__   |
| | |    / _ \ / _` / __\ \ /\ / / _ \| '__| __| '_ \  |
| | |___| (_) | (_| \__ \\ V  V / (_) | |  | |_| | | | |
|  \_____\___/ \__,_|___/ \_/\_/ \___/|_|   \__|_| |_| |
|                                                      |
| Hello! What can I do for you today, master?          |
________________________________________________________
```
* Text based
* Easy to learn
* ~~Better than dog~~

**Features**
- [ ] Remembers your tasks
- [ ] Marks them as completed or not
- [ ] Even keeps track of dates

_All you need to do is,_
1. Download the JAR file [here](https://github.com/wei-song1/ip/releases/tag/A-Jar)
2. Run it
3. Have fun knowing that the future is here!

> [!CAUTION]
> Codsworth is not liable for missed tasks due to user's negligence

This is how you write `Hello World` in Java, because Codsworth loves to say hello to you too!
```
public class Main {
        System.out.println("Hello World");
    }
}
```


src/main/java/codsworth/Task/Task.java
```diff
@@ -0,0 +1,44 @@
+package codsworth.task;
+
+/**
+ * Represents task. A <code>Task</code> object corresponds to a task represented by a name
+ */
+public class Task {
```
cerulyean: Since Task is a base class and shouldn't be instantiated directly, have you considered making it abstract.

src/test/java/codsworth/Task/TaskListTest.java
```diff
@@ -0,0 +1,165 @@
...
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("01-01-2000 012d"));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("01-01-2000 2400"));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("01-01-2000 9999"));
+
+        // Checks for garbage input
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate(""));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate(null));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("a"));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("01-01-2001 abcd"));
+        assertThrows(CodsworthInvalidDateException.class, () -> TaskList.formatCorrectDate("ab-cd-efgh 0123"));
```
cerulyean: I like how this accounts for a very large variety of possible date time formats, it looks very flexible

src/main/java/codsworth/Parser.java
```diff
@@ -0,0 +1,99 @@
+package codsworth;
+
+import codsworth.codsworthexceptions.CodsworthInvalidCommandException;
+import codsworth.codsworthexceptions.CodsworthInvalidDateException;
+import codsworth.codsworthexceptions.CodsworthMissingInputException;
+import codsworth.codsworthexceptions.CodsworthOutOfBoundsException;
+import codsworth.codsworthexceptions.CodsworthWrongFormatException;
```
cerulyean: I like the number of different exceptions, looks like it would be very helpful in maintaining the code and for telling the user what went wrong

src/main/java/codsworth/Task/Task.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        this.taskType = "";
+    }
+
+    public void setDoneOrUndone(String type) {
+        this.isDone = type.equals("mark");
+    }
```
cerulyean: I noticed that the setDoneOrUndone method in the Task class takes a String parameter ("mark" or "unmark") to determine the task’s status. According to coding standards, methods should avoid using strings for control logic, as this can lead to errors and reduce code clarity. Maybe do this under Parser and use a boolean when calling Task methods instead?

src/main/java/codsworth/Task/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+
+            taskList.add(temp);
+            System.out.println(Ui.getNewTask(temp, taskList.size()));
+
+        // Exception
+        } catch (CodsworthMissingInputException e) {
+            throw new CodsworthMissingInputException();
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new CodsworthInvalidDateException();
+        }
```
cerulyean: Catching an exception and throwing a different one seems like it would hide information that is useful. I feel adding more context or just propagating the original exception would be more helpful.

cerulyean (commented):
Your code looks really well organized and neat, with consistently good and descriptive names for methods and variables. The javadoc comments are also descriptive, clear and through. One relatively minor issue(?) I found is you only calling Storage.saveTaskList() when closing the program, but Task-7 states that saving on TaskList modification. Overall, it looks great.

src/main/java/codsworth/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+                System.out.println(e);
+            }
+            break;
+
+        case "bye":
+            isBye = true;
+            Storage.saveTaskList();
+            break;
+
+        case "reset":
```
ZaydM18, line 67: Good choice to add a reset button as the project evolved to include a saved task list.

src/main/java/codsworth/CodsworthExceptions/CodsworthMissingInputException.java
```diff
@@ -0,0 +1,11 @@
+package codsworth.codsworthexceptions;
+
+/**
+ * Thrown when there are missing inputs given to Codsworth
+ */
+public class CodsworthMissingInputException extends RuntimeException {
```
ZaydM18, line 8: Good use of a more complex exception instead of inheriting the base exception class.

src/main/java/codsworth/Task/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+        return output;
+    }
+
+    /**
+     * Adds the correct task to the task list based on the operation given and outputs the corresponding line
+     *
+     * @param input Task name with its optional dates.
+     * @param operation Operation currently being added.
+     */
+    public static void addTaskAndPrint(String input, String operation) {
```
ZaydM18: Seems like this could be compartmentalised into a function for the Parser class instead. Since it looks more like it has to do with instantiation.

src/main/java/codsworth/Task/Task.java
ZaydM18: The Task class is never used directly. Making it abstract would likely be better.

src/main/java/codsworth/Task/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+import codsworth.Ui;
+import codsworth.codsworthexceptions.CodsworthInvalidDateException;
+import codsworth.codsworthexceptions.CodsworthMissingInputException;
+import codsworth.codsworthexceptions.CodsworthOutOfBoundsException;
+import codsworth.codsworthexceptions.CodsworthWrongFormatException;
+
+/**
+ * Represents ArrayList of Task in a TaskList class and encapsulates all it's methods into the class.
+ */
+public class TaskList {
```
ZaydM18, line 14: The extra TaskList class is useful to have, but some of its functions may be performed better by either the Task or Parser classes.

ZaydM18 (commented):
Overall, well-designed code that adheres to OOP principles and demonstrates a good understanding of the core topics. Most issues are minor and can be fixed easily. Some structural modifications may be useful but not necessary. Nothing to worry about, it's no Fallout 76.

PR #227, opened 2025-01-29
[RT0118] iP

RT0118:
**FAUNA**

> "Konfauna~ Your personalised chatbot kirin is here!" - Fauna ([👀](http://www.youtube.com/@niminightmare))

Fauna frees your mind of having to remember the things you need to do in life. It's, 
- text-based
- easy to learn
- _kinda(?)_ fast to use

All you need to do is, 
1. download [Fauna]("https://github.com/RT0118/ip/releases")
2. run it ~~with administrator privileges~~
3. add your tasks
4. profit

Features include:
- [x] Manage tasks, deadlines and events
- [ ] Manage deadlines (not yet)
- [ ] Reminders (not yet)

I still need a code block and `inline code`, so here they are:
```java
    System.out.println("lmao");
}
```


src/main/java/fauna/Fauna.java
```diff
@@ -0,0 +1,163 @@
...
+     * Create a new instance of Fauna
+     * @param saveFileLocation txt save file to load tasks from
+     */
+    public Fauna(String saveFileLocation) {
+        this.ui = new Ui();
+        this.storage = new Storage(saveFileLocation);
+        this.taskList = this.storage.restore();
+    }
+
+    private void addTodoToTaskList(String taskName) {
```
zacklow28: The add__ToTaskList methods can be placed in the TaskList class for better OOP 😃

src/main/java/fauna/parser/FaunaCommand.java
```diff
@@ -0,0 +1,88 @@
+package fauna.parser;
+
+/**
+ * FaunaCommand stores all available commands,
+ * command descriptions and command regex for
+ * extracting variables from user input
+ */
+public enum FaunaCommand {
+    LIST("Lists all tasks"),
```
zacklow28, line 9: Like the ENUMS, I think they really help in code reuse!

src/main/java/fauna/parser/FaunaCommand.java
```diff
@@ -0,0 +1,88 @@
...
+    public String getDescription() {
+        return this.description;
+    }
+
+    /**
+     * <p>Provides a regex pattern string with matching groups
+     * to extract information from user input
+     * </p>
+     * @return regex pattern string
+     */
```
zacklow28, line 76: Interesting use of regular expressions!

src/main/java/fauna/parser/ParsedUserInput.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Optional;
+
+import fauna.exceptions.InvalidUserInputException;
+
+/**
+ * ParsedUserInput stores the values obtained from the user after parsing
+ */
+public class ParsedUserInput {
+    private final FaunaCommand command;
+    private final Optional<String> taskName;
```
zacklow28, line 14: Is the Optional class necessary here? How does it work with your code?

src/main/java/fauna/ui/Ui.java
```diff
@@ -0,0 +1,160 @@
...
+
+import java.util.Scanner;
+
+public class Ui {
+    private static final String LINE_SEPARATOR = "_________________________________________________________________";
+    private static final String CHATBOT_NAME = "Fauna";
+    private static final String CHATBOT_LOGO = """
+             _____ _   _   _ _   _    _    
+            |  ___/ \\ | | | | \\ | |  / \\   
+            | |_ / _ \\| | | |  \\| | / _ \\
```
zacklow28: Wow looks good, good effort! 😮

zacklow28 (commented):
LGTM. Overall demonstrated good use of quite a few Java concepts which I can learn from

src/main/java/fauna/Fauna.java
RT0118: Oh, good suggestion! Thanks for the input

src/main/java/fauna/parser/FaunaCommand.java
RT0118: Thanks for the input!

src/main/java/fauna/parser/ParsedUserInput.java
RT0118: My thought process was to return some kind of meaningful value (which in this case is "invalid ..."), instead of purely returning null. But I do agree, I could have put more thought into using Optionals here.

Correction: I have an OrElseThrow() when you query for something that doesnt exist, not return a string "invalid ..."

src/main/java/fauna/ui/Ui.java
RT0118: Thanks!

src/main/java/fauna/task/DeadlineTask.java
```diff
@@ -0,0 +1,58 @@
...
+        super(deadlineTask, isDone);
+        this.by = deadlineTask.by;
+    }
+
+    public DeadlineTask(String taskName, boolean isDone, LocalDateTime by) {
+        super(taskName, isDone);
+        this.by = by;
+    }
+
+    public DeadlineTask(String taskName, LocalDateTime by) {
```
Kannan171, line 32: Perhaps "by" can be a vague variable name and could be better represented as "deadline" or "doBy"

Kannan171 (commented):
LGTM. Well named and clear to read code

PR #236, opened 2025-01-29
[samuelkwik] iP

samuelkwik:
#### _the okayish chatbot_

> "Only mediocrity can be trusted to be always at its best." – Sir Max Beerbohm

It's not yet a Duke, meet Viscount! For when you already have a calendar app in your pocket but want to make life a little more cumbersome.

Viscount, a blast from the past, is:
- entirely text-based
- ~not really~ intuitive
- okay

All you need to do to get started down a nostalgic trip is,

1. download it [here](https://github.com/samuelkwik/ip)!
2. figure out how to run compile and run it.
3. add your tasks.
4. let it manage it for you the best that it can!

And best of all, it is perpetually **FREE**! ~On the condition that no sane company buys the rights to this and decides that a subscription model is best for users moving forward.~

With a cutting-edge feature list circa 1950,

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Pretending to be your imaginary friend (coming never)

***
If you are a Java programmer (why?), you can use it to practice Java too. Here's the `main` method: 

``` java
    Viscount viscount = new Viscount();
    viscount.run();
}
```

What are you waiting for? Use Viscount now! :sparkles:


src/main/java/viscount/task/Event.java
```diff
@@ -0,0 +1,86 @@
...
+                + seperator + fromDate + seperator + toDate;
+    }
+
+    /**
+     * Returns a string representation of the Event task.
+     *
+     * @return A string representing the Event task, including its type, completion status,
+     * description, and start/end dates.
+     */
+    @Override
```
leewaikin19, line 86: Perhaps it would be useful and easier to use the `@inheritDoc` tag?

src/main/java/viscount/command/DeleteCommand.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String indexStr;
+
+    /**
+     * Constructs a DeleteCommand object with a string representing the index
+     * of the task to be deleted.
+     *
+     * @param indexStr The index of the task to be deleted, provided as a string.
+     */
+    public DeleteCommand(String indexStr) {
+        this.indexStr = indexStr + "";
```
leewaikin19: Is there a need to add `+ "";`?

src/main/java/viscount/command/Parser.java
```diff
@@ -0,0 +1,103 @@
...
+     * @return A Command object representing the user's intended action.
+     * @throws ViscountException If the input string format does not match the intended
+     *                           command
+     */
+    public static Command parse(String inputString) throws ViscountException {
+        return switch (inputString.toLowerCase().split(" ")[0]) {
+            case "todo" -> handleTodo(inputString);
+            case "deadline" -> handleDeadline(inputString);
+            case "event" -> handleEvent(inputString);
+            case "list" -> new ListCommand();
```
leewaikin19, line 99: A suggestion would be to check for parameters after the `list` and `bye` command, to ensure the user does not misunderstand or misuse the command

src/main/java/viscount/command/UnknownCommand.java
```diff
@@ -0,0 +1,23 @@
+package viscount.command;
+
+import viscount.Storage;
+import viscount.task.TaskList;
+import viscount.TextUi;
+
+/**
+ * Represents a command that handles invalid or unrecognized user input.
+ */
+public class UnknownCommand extends Command {
```
leewaikin19, line 10: Great use of a separate `UnknownCommand` class to handle invalid cases!

src/main/java/viscount/task/Event.java
samuelkwik: Thank you for this! I honestly did not know that the `@inheritDoc` tag exists... you can probably imagine the pain it was to ensure consistency amongst all the child classes! Thanks again! 😄

src/main/java/viscount/command/DeleteCommand.java
samuelkwik: Thank you for pointing that out! It is definitely not supposed to be there.

src/main/java/viscount/command/Parser.java
samuelkwik: Good suggestion! Thank you!

src/main/java/viscount/command/UnknownCommand.java
samuelkwik: 😸 Thank you!

PR #243, opened 2025-01-29
[Kannan171] iP

Kannan171:
### Chitti

> “One of the secrets of getting more done is to make a TO-DO List every day, keep it visible, and use it as a guide to action as you go through the day.” — Jean de La Fontaine

Chitti is a todo list manager for you, _its_ 
- text based
- easy to use

All you need to do is
1. Download it from [here](https://github.com/Kannan171/ip)
2. Start adding your tasks 😊

Have **YOU** started using Chitti?
- [ ] Downloaded and started

Here is the Java code to get you started on launching :
`public class Main {
    }
}`


src/main/java/Chitti.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final String FILE_PATH = "./chitti.txt";
+    public static void main(String[] args) {
+
+        System.out.println("Hello! I'm Chitti");
+        System.out.println("What can I do for you?");
+        Scanner s = new Scanner(System.in);
+        ArrayList<Task> taskList = loadTasksFromFile();
+        String input = s.nextLine();
+        while (!input.equals("bye")){
+            if (input.equals("list")) {
```
feliciahmq: Perhaps we can use switch case here instead of if else?

src/main/java/Chitti.java
```diff
@@ -0,0 +1,110 @@
...
+                int ind = Integer.parseInt(input.substring(7));
+                Task toRemove = taskList.get(ind-1);
+                taskList.remove(ind-1);
+                System.out.println("Noted. I've removed this task:");
+                System.out.println(toRemove);
+                System.out.println("Now you have " + taskList.size() + " tasks in the list.");
+            } else {
+                try {
+                    Task newTask;
+                    if (input.startsWith("deadline")) {
```
feliciahmq: Maybe we can have a general method eg. execute() for Command, then allow each subclasses to override it with task-specific logic.

src/main/java/Chitti.java
```diff
@@ -0,0 +1,110 @@
...
+        Scanner s = new Scanner(System.in);
+        ArrayList<Task> taskList = loadTasksFromFile();
+        String input = s.nextLine();
+        while (!input.equals("bye")){
+            if (input.equals("list")) {
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 0; i < taskList.size(); i++) {
+                    System.out.println((i + 1) + "." + taskList.get(i));
+                }
+            } else if (input.startsWith("mark")){
```
feliciahmq: Consider extracting the different commands into a general abstract Command class.

src/main/java/Task.java
```diff
@@ -0,0 +1,43 @@
+public class Task {
```
feliciahmq: I like your OOP approach here.

feliciahmq (commented):
LGTM 👍🏻

src/main/java/chitti/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class Parser {
+    /**
+     * Handles the different commands that users can input
+     * @param input
+     * @param taskList
+     * @param ui
+     */
+    void handleCommand(String input, TaskList taskList, Ui ui) {
+        if (input.equals("list")) {
```
Parrycys: Perhaps we could use switch case statements here ?

src/main/java/chitti/Task.java
```diff
@@ -0,0 +1,49 @@
+package chitti;
+
+public class Task {
+    private boolean isDone;
+    private String name;
+    public Task(String name){
```
Parrycys, line 18: Perhaps a new line after declaring variables/before constructor ?

src/main/java/chitti/Ui.java
```diff
@@ -0,0 +1,57 @@
...
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < tasks.size(); i++) {
+            System.out.println((i + 1) + "." + tasks.get(i));
+        }
+    }
+
+    void showFoundList(ArrayList<Task> tasks){
+        System.out.println("Here are the matching tasks in your list:");
+        for (int i = 0; i < tasks.size(); i++){
+            System.out.println((i+1)+"." + tasks.get(i));
```
Parrycys: Maybe we can add whitespaces between the 2nd "+" symbol ?

Parrycys (commented):
LGTM. 😃 I like how your code is very neat and well indented !

PR #247, opened 2025-01-29
[GaaneshT] iP

GaaneshT:
### Malt

> “Your mind is for having ideas, not holding them.” – David Allen [(source)](source_link_here)

DukePro frees your mind of having to remember things you need to do. It's,

- text-based  
- easy to learn  
- **FAST _SUPER_ FAST** to use  

All you need to do is,

1. download it from [here](download_link_here).  
2. double-click it.  
3. add your tasks.  

And it is **FREE!**  

### Features:
- [x] Managing tasks  
- [x] Managing deadlines (coming soon)  
- [ ] Reminders (coming soon)  

```java
public class Main {
    }
}
```


src/main/java/malt/parser/Parser.java
```diff
@@ -0,0 +1,224 @@
...
+    public static boolean parseAndExecute(String input, TaskList tasks, Ui ui, Storage storage) throws MaltException {
+        // Split the first word as 'command', the rest as 'arguments'
+        String[] parts = input.split(" ", 2);
+        String command = parts[0].toLowerCase();
+        String argument = (parts.length > 1) ? parts[1].trim() : "";
+
+        switch (command) {
+            case "bye":
+                // User wants to exit
+                return handleBye(ui);
```
Jeremiah33-3: just a suggestion! perhaps the naming of the methods can be made clearer if the word `Command` is specified.
```java
...
return handleByeCommand(ui);
...
return handleListCommand(ui);
...
```

src/main/java/malt/parser/Parser.java
```diff
@@ -0,0 +1,224 @@
...
+                throw new MaltException("I'm sorry, but I don't know what that means!");
+        }
+    }
+
+    // --------------------------------------------------
+    // COMMAND HANDLERS
+    // --------------------------------------------------
+
+    private static boolean handleBye(Ui ui) {
+        ui.showGoodbye();
```
Jeremiah33-3, line 116: Similarly, might be easier for readers to understand if this method name is changed to:
```java
showGoodbyeMessage();
```

src/main/java/malt/parser/Parser.java
```diff
@@ -0,0 +1,224 @@
...
+            storage.saveTasks(tasks.getAllTasks());
+        } catch (NumberFormatException e) {
+            throw new MaltException(" Invalid index for unmark command!");
+        }
+    }
+
+    private static void handleDelete(String arg, TaskList tasks, Ui ui, Storage storage) throws MaltException {
+        try {
+            int index = Integer.parseInt(arg);
+            Task removed = tasks.removeTask(index - 1);
```
Jeremiah33-3: Here, `removed` might be mildly ambiguous, `removedTask` might be easier to understand!

src/main/java/malt/parser/Parser.java
```diff
@@ -0,0 +1,224 @@
...
+        ui.showLine();
+        System.out.println(" Adding this task:");
+        System.out.println("   " + deadline);
+        System.out.println(" Now you have " + tasks.size() + " tasks in the list! Get working :(");
+        ui.showLine();
+    }
+
+    private static void handleEvent(String arg, TaskList tasks, Ui ui, Storage storage) throws MaltException {
+        // e.g.: "project meeting /from Monday 2pm /to 4pm"
+        String[] fromSplit = arg.split("/from", 2);
```
Jeremiah33-3: emmm `fromSplit` sounds a bit confusing as to what this array is supposed to represent! might there be a clearer naming?

src/main/java/malt/ui/Ui.java
```diff
@@ -0,0 +1,105 @@
...
+     * Displays the Malt ASCII art logo.
+     */
+    private static void displayLogo() {
+        String logo = """
+                          _____                    _____                    _____        _____         \s
+                         /\\    \\                  /\\    \\                  /\\    \\      /\\    \\        \s
+                        /::\\____\\                /::\\    \\                /::\\____\\    /::\\    \\       \s
+                       /::::|   |               /::::\\    \\              /:::/    /    \\:::\\    \\      \s
+                      /:::::|   |              /::::::\\    \\            /:::/    /      \\:::\\    \\     \s
+                     /::::::|   |             /:::/\\:::\\    \\          /:::/    /        \\:::\\    \\    \s
```
Jeremiah33-3, line 24: this is actually so cool 😮

Jeremiah33-3 (commented):
Overall, good work!

src/main/java/malt/parser/Parser.java
GaaneshT: Thanks for suggestion, I'll be sure to amend it!

src/main/java/malt/parser/Parser.java
GaaneshT: Good point, thanks!

src/main/java/malt/parser/Parser.java
GaaneshT: I agree, will change it. Thanks!

src/main/java/malt/ui/Ui.java
GaaneshT: :D

src/main/java/malt/parser/Parser.java
GaaneshT: I thought fromSplit would be good since it like splits it at the "from" field so hence fromSplit. But you make a good point and I can see how it would be confusing. Will fix it

PR #253, opened 2025-01-29
[Taoseeker] iP

Taoseeker:
# NeoChat
>“Smart people have already started using honorifics when talking to AI.”  - Me

Neochat frees your mind of having to remember things you need to do. 
It's,

- text-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use

All you need to do is

1. download it from [here](https://github.com/Taoseeker/ip/releases/tag/A-Jar)
2. Turn on intelliJ idea
3. make sure the `.jar` file is in working directory
4. in your terminal, run the command `java -jar "neochat.jar"`
5. add your tasks
6. let it manage your tasks for you 😉

And it's **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [ ] AI assistant (coming sooooooon)


src/main/java/neochat/task/Parser.java
```diff
@@ -0,0 +1,152 @@
...
+                printCommandList();
+                break;
+            case "find":
+                taskList.findTasks(remainingInput);
+                break;
+            default:
+                System.out.println("Unknown command type" + commandType);
+            }
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
```
Nano-233, line 90: would a error log or user-friendly message be better than printing the internal details? for example e.getmessage

src/main/java/neochat/task/Parser.java
```diff
@@ -0,0 +1,152 @@
...
+            this.taskList.addTask(new Todo(input.trim()));
+        } catch (EmptyTaskDescriptionException e) {
+            System.out.println("Task description is empty");
+        }
+    }
+
+    private void parseDeadline(String input) {
+        String[] parts = input.split(" /by ", 2);
+        if (parts.length < 2) {
+            throw new IllegalArgumentException("Deadline format error，the correct format should be" +
```
Nano-233: Is there a reason that a unstandardized comma was used here instead of the standard comma? I noticed it in a few other places too

src/main/java/neochat/task/Parser.java
```diff
@@ -0,0 +1,152 @@
...
+        System.out.println("1. list - Show all tasks");
+        System.out.println("2. todo <description> - Add a Todo task");
+        System.out.println("3. deadline <description> /by <yyyy-MM-dd HH:mm> - Add a Deadline task");
+        System.out.println("4. event <description> /from <yyyy-MM-dd HH:mm> /to <yyyy-MM-dd HH:mm> - Add an Event task");
+        System.out.println("5. mark <task number> - Mark a task as done");
+        System.out.println("6. unmark <task number> - Mark a task as not done yet");
+        System.out.println("7. help - Show the command list");
+        System.out.println("8. bye - Exit the program");
+        System.out.println("____________________________________________________________");
+    }
```
Nano-233: Perhaps a separate utility class is most suitable for this static method as it is unrelated to parsing?

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
+package neochat.task.tasklist;
+
+import java.io.*;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.*;
+
+import neochat.task.*;
```
Nano-233: Perhaps non-wildcard imports would be more suitable for our coding standards? I assume IDE issue since other ones are fine?

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+    private final File savedListFile;
+
+
+    /**
+     * Constructs a TaskList and initializes it by loading tasks from the saved file.
+     * Creates the file and necessary directories if they don't exist.
+     */
+    public TaskList() {
+        this.tasks = new ArrayList<>(100);
+        this.savedListFile = new File("src/data/savedList.txt");
```
Nano-233, line 43: Perhaps using a constant could be more suitable for file handling to allow for easier modifications and prevent hard-coded path related errors?

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+    /**
+     * Parses a string from the save file into a Task object.
+     *
+     * @param line The saved task string in format "[Type] | [Status] | [Description] | [DateTime]"
+     * @return Reconstructed Task object, or null if format is invalid
+     */
+    private Task parseTask(String line) {
+        String[] parts = line.split(" \\| ");
+        if (parts.length < 3) {
+            return null;
```
Nano-233, line 132: perhaps a log of invalid format would be useful for debugging (as there is an invalid line)

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+     */
+    public void printList() {
+        if (count == 0) {
+            System.out.println("Empty task list!");
+        } else {
+            for (int i = 0; i < count; i++) {
+                Task task = tasks.get(i);
+                System.out.println((i + 1) + ": " + task.toString());
+            }
+        }
```
Nano-233: is there a reason a global count is used instead of for example task.size() ?

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+            System.out.println("____________________________________________________________");
+            System.out.println("OK, I've marked this task as not done yet:");
+            System.out.println("  " + task);
+            System.out.println("____________________________________________________________");
+        } catch (NumberFormatException e) {
+            System.out.println("Invalid input. Please provide a valid task number.");
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("Invalid task number. Please provide a number between 1 and " + count + ".");
+        }
+    }
```
Nano-233: Perhaps could combine the two methods as they seem to share many lines of code?

Nano-233 (commented):
LGTM! Just small alternatives that perhaps could be considered to make the code easier for maintaining and reading!

src/main/java/neochat/task/Parser.java
Taoseeker: Thank you for your suggestion! I was thinking using this could help me debug. Will make the error message more user-friendly in future release.

src/main/java/neochat/task/tasklist/TaskList.java
Taoseeker: Thanks for pointing it out. Will use explicit import in future.

src/main/java/neochat/task/Parser.java
Taoseeker: No specific reason, it is just my standard keyboard input is Chinese(I wrote short comment in Chinese sometime for my own reference) so sometimes I mistakenly use Chinese comma.

src/main/java/neochat/task/Parser.java
Taoseeker: You do have a valid point. I plan to make Ui class handle this in future release.

src/main/java/neochat/task/tasklist/TaskList.java
Taoseeker: I wrote my own data structure in an early version which does not keep track of the count, so this global count is 'inherited' from earlier version and I forgot to edit.

src/main/java/neochat/task/tasklist/TaskList.java
Taoseeker: I will consider it.

src/main/java/neochat/task/tasklist/TaskList.java
Taoseeker: You are right: my initial thought is to let caller of `parseTask` to handle null, but I forgot to implement it. Will do so in future version.

src/main/java/neochat/task/tasklist/TaskList.java
```diff
@@ -0,0 +1,295 @@
...
+        } catch (IOException e) {
+            System.out.println("Fail to save tasks" + e.getMessage());
+        }
+    }
+
+    private static LocalDateTime parseDateTime(String dateTimeStr) {
+        try {
+            return LocalDateTime.parse(dateTimeStr, DATE_TIME_FORMATTER);
+        } catch (DateTimeParseException e) {
+            throw new IllegalArgumentException("Invalid time format, the correct format should be "
```
weiwong834, line 230: Would a checked exception such as DateTimeFormatException be better in this case?

src/main/java/neochat/task/tasklist/TaskList.java
weiwong834: Maybe the all the println lines can be handled in the Ui class to simplify these methods.

src/main/java/neochat/ui/Ui.java
```diff
@@ -0,0 +1,20 @@
+package neochat.ui;
+
+import java.util.*;
```
weiwong834: Maybe consider not using wildcard imports

weiwong834 (commented):
Overall looks good! Perhaps a shifting some of the implementations into Ui class will improve clarity.

PR #254, opened 2025-01-29
Hsin Yi iP

hsinyilow:
### Astra :sparkles:

Astra helps you keep track of the many ideas you may have.
It is
- _fast_
- lightweight
- **free**

To use it, you just need to:
1. Download it from [here](https://github.com/hsinyilow/ip/)
2. Double click it
3. Add your tasks

Features:
- [x] Marking tasks
- [x] Shows date for deadline and events
- [ ] Manages deadlines

You can use it to practice Java too! Here is a snippet of the `command` method
``` ruby
    public void command(String input) {
        if (input.equals("list")) {
            //show list
            displayTaskList();
        } else if (input.startsWith("delete")) {
            deleteTask(input);
        }
        ...
    }
```


src/main/java/astra/Astra.java.rej
kubrian, line 1: I think this file [should be removed](https://stackoverflow.com/questions/34585865/what-are-rej-files-which-are-created-during-merge)

src/test/java/astra/system/ParserTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+public class ParserTest {
+
+    @Test
+    public void testCommand1() {
+        assertEquals("aee", Parser.parseCommand("       a     e     e   ", 2, true));
+    }
```
kubrian: I think a more descriptive name, as well as a comment to explain the string will make the purpose of test case more intuitive.

src/main/java/astra/task/Task.java
```diff
@@ -0,0 +1,44 @@
+package astra.task;
+
+/**
+ * Is inherited by all tasks.
```
kubrian: Consider using "Represents an abstract task that ..."

kubrian (commented):
LGTM

src/main/java/astra/Astra.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public Astra(String filePath) {
+        SaveSystem.loadSaveFile(filePath, taskList);
+        Ui.greet();
+    }
+
+    /**
+     * Command and result loop
+     */
+    public void runAstra() {
```
DanielJames0302, line 32: Good method naming

src/main/java/astra/system/Parser.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public static TimeData parseTime(String input) throws AstraException {
+        String[] parseInput = input.split(" ");
+
+        try {
+            LocalDate date = LocalDate.parse(parseInput[0]);
+            LocalTime time = LocalTime.MIN;
+            if (parseInput.length != 1) {
+                time = LocalTime.of(Integer.parseInt(parseInput[1].substring(0, 2)),
+                        Integer.parseInt(parseInput[1].substring(2)));
```
DanielJames0302: Well aware of the indentation

DanielJames0302 (commented):
LGTM. Well done.

PR #260, opened 2025-01-30
[mjyan02] iP

mjyan02:
# Duke Chatbot

Duke frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is:

1. download it from [here](https://github.com/mjyan02/ip.git).
2. double-click it.
3. add your tasks.

And it is **FREE**!

**Features:**

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

> [!NOTE] If you are a Java programmer, you can use it to practice Java too. Here's the `main` method:

```
public class Main {
        Application.launch(DukeLauncher.class, args);
    }
}
```


src/main/java/Duke/Task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Converts the task into a formatted string representation suitable for file storage on hard disk.
+     *
+     * @return A string representation of the task for file storage.
+     */
+    @Override
+    public String toFile() {
+        return "D | " + (isDone ? "1" : "0") + " | " + description + " | " + by;
```
chinpcbenjamin: Is it better to separate this into multiple lines? It could improve code readability by using having each "+" term on a new line.

Alternatively, String.format could be useful too! 😄 

```suggestion
        return "D | "
            + (isDone ? "1" : "0")
            + " | "
            + description
            + " | "
            + by;
```

or even:

```suggestion
        return String.format("D | %d | %s | %s", isDone ? 1 : 0, description, by);
```

src/main/java/Duke/Task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a task with a description and a due date.
+ */
+public class Deadline extends Task {
+    /**
+     * The due date of the task.
+     */
+    protected LocalDate by;
```
chinpcbenjamin: Would it be better to use another name for this? The word "by" does not tell me a lot about what the variable is for. Perhaps it could be changed to be another name such as "dueDate" 👍

src/main/java/Duke/Task/Event.java
```diff
@@ -0,0 +1,82 @@
...
+public class Event extends Task {
+    /**
+     * The start date of the event.
+     */
+    protected LocalDate from;
+
+    /**
+     * The end date of the event.
+     */
+    protected LocalDate to;
```
chinpcbenjamin: Would it be better to use another name for these? The words "by" and "from" do not tell me a lot about what the variables iare for. Perhaps it could be changed to be another name such as "dueDate" and "beginDate". 👍

chinpcbenjamin (commented):
Looks good 👍 . The code was easy to read, perhaps except for the various toString methods.

src/main/java/Duke/Exceptions/DukeException.java
```diff
@@ -0,0 +1,15 @@
+package duke.exceptions;
```
reaneechua: Is this class necessary? Exception on its own may be sufficient already.

src/main/java/Duke/Task/Keywords.java
```diff
@@ -0,0 +1,70 @@
+package duke.task;
```
reaneechua: I liked the way this class was created just to keep the keywords as enums, helps to mitigate user errors.

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,96 @@
...
+    public void run(String input) {
+        try {
+            if (input.equalsIgnoreCase("help")) {
+                ui.welcomeMessage();
+                return;
+            }
+
+            Keywords keyword = Parser.parseKeyword(input);
+
+            switch (keyword) {
```
reaneechua: Maybe this logic could be passed to the Parser class instead? Seems like there is abit of passing back and forth between the main class and Parser, it might be possible for Parser to handle this logic.

reaneechua (commented):
Overall, I think your code was well-written and logic was sound. I didn't spot any coding violations. I feel that the class structures could be improved, but you did great! Keep up the good work!

PR #264, opened 2025-01-30
[FAN YUEH-CHING] iP

e1093051:
### **Carolyn**

> I must govern the clock not govern by it [-Golda Meir](https://woodenswisdom.com/leadershipIssue.php?issue_id=1244)

This app allows you

- mark and unmark tasks
- add new tasks

All you need to do is

1. Download it
2. **Run it!!**

Features

- [ ] Manage task
- [ ] Manage deadline😎

Compile the program with
`javac Carolyn.java`

The below is the main class
```java
public class Main {
    }
};
```


src/main/java/carolyn/Ui.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.util.Scanner;
+
+public class Ui {
+    private Scanner scanner;
+    public Ui() {
+        this.scanner = new Scanner(System.in);
+    }
+  
+    String INDENT = "    ";
```
SuspectBlue: I believe you intent for INDENT to be a constant. In that case, you should add the static and final keywords to it.

src/main/java/carolyn/CarolynException.java
SuspectBlue, line 1: Consider using more specific exception types instead of a single general exception for more precise error handling.

src/main/java/carolyn/Task.java
SuspectBlue, line 1: Consider organizing similar classes into one folder to split them up in packages. For example you can place the Task.java, Event.java, Deadline.java, Todo.java and TaskList.java all in a task folder.

SuspectBlue (commented):
LGTM!
Just some minor nits.🙂

PR #281, opened 2025-01-30
[Victoria281] iP

Victoria281:
# Vic

> "Your mind is for having ideas, not holding them." – David Allen [(source)](https://gettingthingsdone.com/)

**Vic** frees your mind from remembering tasks and helps you stay organized. It's:

- **Text-based**
- **Easy to learn**
- ~~**FAST**~~ ⚡ **SUPER FAST** to use

All you need to do is:

1. [Download it from here](#).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you! 😉

And the best part? **It's FREE!**

## Features

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon!)

If you are a Java programmer, you can use **Vic** to practice Java too. Here's the main method:

```java
public class Main {
    }
}
```

Start using **Vic** today and experience the power of efficient task management! 🚀


src/main/java/Vic.java
```diff
@@ -0,0 +1,237 @@
...
+                if (description.length() < 1) throw new EmptyContentException();
+
+                String by = response.split("/by ").length > 1 ? response.split("/by ")[1] : "";
+                if (by.length() < 1) throw new EmptyContentException();
+
+                newItem = new Deadline(description, by);
+                break;
+            case EVENT:
+                description = response.split(" ", 2)[1].split("/")[0];
+                if (description.length() < 1) throw new EmptyContentException();
```
pang-yy: I think it would be better to add braces `{}` for `if` statement.

src/main/java/Vic.java
```diff
@@ -0,0 +1,237 @@
...
+
+                String[] part2 = response.split("/");
+                String from = part2.length > 1 ? part2[1].split("from ").length > 1 ? part2[1].split("from ")[1] : "" : "";
+                if (from.length() < 1) throw new EmptyContentException();
```
pang-yy: I think it would be better to add braces `{}` for `if` statement.
I noticed the same issue in several other places too.

src/main/java/Vic.java
```diff
@@ -0,0 +1,237 @@
...
+     *
+     * @param option task id provided by user
+     * @return task id or -1 for error
+     */
+    static int parseTaskId(String option) {
+        int taskID = -1;
+        try {
+            taskID = Integer.parseInt(option);
+        }
+        catch(NumberFormatException e) {
```
pang-yy: I think `catch(...) {` should be on the same line as `}`.
I noticed the same issue in several other places too.

src/main/java/Vic.java
```diff
@@ -0,0 +1,237 @@
...
+    static void deleteListItem(String response) {
+        String[] responseLst = response.split(" ");
+        try {
+            if (responseLst.length <= 1) throw new EmptyContentException();
+        } catch (EmptyContentException e) {
+            System.out.println(replyFormatter(e.getMessage()));
+            return;
+        }
+        int taskID = parseTaskId(responseLst[1]);
+        if (taskID == -1) { return; }
```
pang-yy: I think `{` should have line break after it.

pang-yy (commented):
Overall, I found your code easy to read for the most part, except a few places where the coding standard was not followed.
Good work! 👍

src/main/java/Vic.java
DanielJames0302: Perhaps follow Egyptian braces ?

src/main/java/Vic.java
DanielJames0302: Perhaps should follow Egyptain braces

src/main/java/Vic.java
```diff
@@ -0,0 +1,237 @@
...
+     */
+    static void addListItem(String response) {
+        String[] responseLst = response.split(" ");
+        Task newItem = null;
+        String description = "";
+        try {
+            if (responseLst.length <= 1) throw new EmptyContentException();
+            Command command = Command.convertText(responseLst[0]);
+            switch (command) {
+            case TODO:
```
DanielJames0302: Well aware of the indentation of switch case

DanielJames0302 (commented):
LGTM. Just some nits to fix.

src/main/java/Vic.java
Victoria281: Thanks for the reminder!

src/main/java/Vic.java
Victoria281: Thanks for the reminder!

src/main/java/Vic.java
Victoria281: I will take note. Thank you!

src/main/java/Vic.java
Victoria281: I missed that out! Thanks

Victoria281: Thanks for the reminder!

src/main/java/Vic.java
Victoria281: Thanks for the reminder!

PR #290, opened 2025-01-30
[bhavinaa] IP

bhavinaa:
# 🚀 Bhavs  

> “Lost time is never found again” – Benjamin Franklin.  

## 📌 What is Bhavs?  
Bhavs is a **task management and productivity tool** designed to help users stay organized, meet deadlines, and efficiently track their goals. Whether you're a student, professional, or entrepreneur, Bhavs provides an intuitive interface to manage tasks seamlessly.  

## Overview  
Bhavs is a lightweight, **text-based** productivity application 🚀 that allows users to:  
- **Add and manage tasks** effortlessly  
- **Track deadlines** to never miss important dates  
- **Set reminders** for upcoming tasks  
- **Filter tasks** based on priority, due date, or completion status  

## Motivation  
The idea behind Bhavs was born out of the **need for a simple yet effective** task management solution. Many existing tools are either too complex or lack the *flexibility* needed for different workflows.  

This is a ~~simple~~ very simple markdown tool.  

Our key motivations:  
1. **Simplicity & Usability** – Creating an easy-to-use tool without unnecessary complexity  
2. **Productivity & Focus** – Helping users organize their tasks without distractions  
3. **Efficiency & Speed** – A lightweight and fast application for quick task management  

## 🛠 Features  
- ✅ Task management  
- ✅ Deadline tracking  
- ✅ Reminders  
- ✅ Task filtering  

## Sample Code  
```java
public class Main {
        System.out.println("Hello, world!");
    }
}
```

## Author

[More about the Author](https://github.com/bhavinaa)


src/main/java/bhavs/tasks/Deadlines.java
```diff
@@ -0,0 +1,29 @@
+package bhavs.tasks;
+
+import bhavs.utils.Time;
+
+public class Deadlines extends Task {
```
lesterlimjj, line 7: Should we name the class as Deadlines when we are creating a single Deadline?

src/main/java/bhavs/tasks/Events.java
```diff
@@ -0,0 +1,40 @@
...
+
+import bhavs.utils.Time;
+
+/**
+ * A type of tasks that includes two start and end times
+ * two constructors is needed to initalise this as the format of the input from hte user is different from
+ * the way it is read from the file
+ **/
+
+public class Events extends Task {
```
lesterlimjj, line 10: Same comment as the one given to deadline

src/main/java/bhavs/tasks/ToDos.java
```diff
@@ -0,0 +1,22 @@
+package bhavs.tasks;
+
+public class ToDos extends Task {
```
lesterlimjj, line 5: Same comment as the one given to deadline

src/main/java/bhavs/tasks/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.Scanner;
+
+/**
+ * Manages a list of tasks, including adding, deleting, marking, and unmarking tasks.
+ * This class interacts with the {@code Storage} class to save task changes.
+ */
+public class TaskList {
+
+    private List<Task> taskList = new ArrayList<>();
+    public Storage storedIn;
```
lesterlimjj: Class variables should not be declared as public. Perhaps we can have storage be a field in main and perform its function calls there instead?

lesterlimjj: Also the name storedIn implies that this variable tells where the data is store (e.g. a file path) which can be confusing. Giving it a name such as store or fileStore would be more appropriate.

src/main/java/bhavs/utils/Time.java
```diff
@@ -0,0 +1,67 @@
+package bhavs.utils;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Scanner;
+
+public class Time {
```
lesterlimjj, line 12: I like the idea of a Time class!

src/main/java/bhavs/utils/UI.java
```diff
@@ -0,0 +1,155 @@
+package bhavs.utils;
+
+import bhavs.tasks.*;
+import java.util.Scanner;
+
+/*
+the UI class contains all the helper functions that allows you to interact with
+the system and access items from the storage
+*/
+public class UI {
```
lesterlimjj, line 10: Your UI class is handling more responsibilities than just displaying UI elements. Consider extracting out these extra responsibilities to other classes. 

A suggestion would be to have your storage, scanner, parser and taskList fields be in main instead and perform their work there.

src/main/java/bhavs/utils/UI.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    public String getUserInput(Scanner scanner, String prompt) {
+        if (prompt != null) {
+            System.out.println(prompt);
+        }
+        return scanner.nextLine();
+    }
+
+    public void processComands() {
```
lesterlimjj: Perhaps this could be in Parser instead, since it already handles commands related to task creation?

src/main/java/bhavs/utils/UI.java
```diff
@@ -0,0 +1,155 @@
...
+            System.out.println("Your task list is empty.");
+        } else {
+            System.out.println("Here are your tasks:");
+            for (int i = 0; i < taskList.size(); i++) {
+                System.out.println((i + 1) + ". " + taskList.get(i));
+            }
+        }
+    }
+
+    public Task make_correct_entry(String userCommand) {
```
lesterlimjj: Your function names should be in camelCase.

src/main/java/bhavs/tasks/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+     * @param scanner
+     *         Scanner object used to get user input.
+     * @param prompt
+     *         The prompt message displayed to the user.
+     * @param listSize
+     *         The number of tasks in the list.
+     *
+     * @return The valid task index.
+     */
+    public int getTaskIndex(Scanner scanner, String prompt, int listSize) {
```
lesterlimjj: It's a cool design choice to have your user be prompted for index until a correct one is given when CRUD operations(e.g. mark, delete). However, if a user were to input it by mistake, they have no way of cancelling the command except for inputting a valid index.

src/main/java/bhavs/utils/Storage.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.*;
+
+/**
+ * Handles loading and saving of tasks to a file.
+ * Ensures that tasks persist between program runs.
+ */
+public class Storage {
+
+    private String filePath;
+    public TaskList taskList;
```
lesterlimjj: Class fields should not be public. Same suggestion of moving these field to main and handling the work there is given.

src/main/java/bhavs/utils/Time.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final DateTimeFormatter INPUT_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+    private static final DateTimeFormatter OUTPUT_FORMAT = DateTimeFormatter.ofPattern("MMM dd yyyy, hh:mm a");
+    private LocalDateTime dateTime;
+
+    public Time(String input, boolean isInteractive) {
```
lesterlimjj, line 31: Extending the function of a constructor to check for prompt or parsing is interesting but not really the intended use of a constructor. Perhaps this logic can be handled elsewhere?

lesterlimjj (commented):
I would like to mention that this is my 2nd PR review so I am not putting much focus on the coding standard, but instead code quality (with a focus on naming). The code here has some interesting and cool design choice (e,g, prompting user for a correct input after a command is locked in) but needs to be careful of over extending the functionality classes.

src/main/java/bhavs/utils/UI.java
jeronkk: This is a typo, it should be processCommands().

src/main/java/bhavs/tasks/TaskList.java
```diff
@@ -0,0 +1,136 @@
...
+     * @param scanner
+     *         Scanner object used to get user input.
+     */
+    public void unmarkTask(Scanner scanner) {
+        if (taskList.isEmpty()) {
+            System.out.println("No tasks to unmark.");
+            return;
+        }
+
+        int taskIndex = getTaskIndex(scanner, "Which task number to unmark?", taskList.size());
```
jeronkk: Both markTask and unmarkTask have similar logic, maybe implementing a helper method to reduce the number of lines?

src/main/java/bhavs/utils/Parser.java
```diff
@@ -0,0 +1,44 @@
+package bhavs.utils;
+
+import bhavs.tasks.Deadlines;
+import bhavs.tasks.Events;
+import bhavs.tasks.*;
```
jeronkk: Best is to not use Wildcard (*) imports as it can lead to ambiguity and unnecessary imports.

src/main/java/bhavs/tasks/Task.java
```diff
@@ -0,0 +1,74 @@
...
+        this.isDone = true;
+    }
+
+    public void markAsIncomplete() {
+        this.isDone = false;
+    }
+
+    public boolean isCompleted() {
+        return isDone;
+    }
```
jeronkk, line 57: There are some missing javadoc comments here.

src/main/java/bhavs/utils/Time.java
jeronkk, line 12: Nice use of the Time class to add more OOP to your application

jeronkk (commented):
LGTM. Just some nits to fix.

PR #307, opened 2025-01-30
[amir-dha] iP

amir-dha:
# Mochi
> **_Bullying you into success, one task at a time._**
 
###  _Say hi to Mochi, who ironically, is not sweet at all and **WILL** badger you to be on task. It's giving:_
- :family_man_woman_girl_boy: Kiasu Parent Vibes (KPV)
- :+1: Very gwenchana to use
- :computer: An online chingu, no cap

### _Well, welcome to your first task:_
- [ ] Download the latest version [here](https://github.com/amir-dha/ip/releases/download/A-Jar/mochi.jar).
- [ ] Run the JAR file.
- [ ] Start adding things you are totally "going" to do.

## Features

1. :calendar: Task Organization, because we should.
2. :hourglass: Deadlines & Events tracking, because we hope.
3. :fire: Persistent file-based storage, because we care.
4. :microphone: A lovely yapper, because we can.

### Now, if you are Java bro, you can use it to practise Java leh. Here, got main method:
```
public class Main {
    }
}
```


src/main/java/mochi/command/Command.java
```diff
@@ -0,0 +1,158 @@
+package mochi.command;
+
+
+import mochi.task.Task;
+import mochi.task.Todo;
+import mochi.task.Deadline;
+import mochi.task.Event;
+import mochi.task.TaskList;
```
NightlyAffair: I think in the case of more than 4 imports, it is acceptable to use wildcard imports. Noticed this in other files as well.

`import mochi.task.*`

src/main/java/mochi/command/Parser.java
NightlyAffair, line 1: Personally love the personality in the interactions with users

src/main/java/mochi/command/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+public class Parser {
+
+    /**
+     * Parses a user command from input.
+     * @param input The raw user input.
+     * @return A command instance representing the user's input.
+     * @throws MochiException MochiException if the input is invalid.
+     */
+    public static Command parse(String input) throws MochiException {
+        String[] pts = input.split(" ", 2);
```
NightlyAffair: Perhaps pts could be named better maybe userCommand?

src/main/java/mochi/storage/Storage.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Parses a task from a formatted string.
+     * @param line The formatted task string.
+     * @return The parsed Task object, or null if parsing fails.
+     */
+    private Task parseTask(String line) {
+        try {
+            String[] parts = line.split(" \\| ");
+            String type = parts[0];
```
NightlyAffair, line 81: Could use ENUMs for the type of user command as it is a constant allowing for easier naming

NightlyAffair (commented):
I think most variables are well named. However i feel that there is a tendency for shorter names where longer names would aid in understanding and readability.

src/main/java/mochi/Mochi.java
```diff
@@ -0,0 +1,76 @@
...
+        ui.showWelcome();
+        boolean isBye = false;
+        Scanner scanner = new Scanner(System.in);
+
+        while (!isBye) {
+            try {
+                String instruction = scanner.nextLine().trim();
+                Command command = Parser.parse(instruction);
+                command.exec(tasks, ui, storage);
+                isBye = command.isBye();
```
Kannan171: LGTM, well abstracted, code is very neat

src/main/java/mochi/task/Task.java
```diff
@@ -0,0 +1,55 @@
+package mochi.task;
+
+/**
+ * Represents a general task with a description and completion status.
+ */
+public abstract class Task {
```
Kannan171, line 6: I like your OOP approach to the tasks

src/main/java/mochi/command/Command.java
```diff
@@ -0,0 +1,158 @@
...
+import mochi.storage.Storage;
+import mochi.ui.Ui;
+import mochi.MochiException;
+import java.io.IOException;
+
+/**
+ * Represents an abstract command that can be executed on the task list.
+ */
+public abstract class Command {
```
Kannan171, line 17: Instead of having so many different classes, perhaps it may be better to have them as separate methods instead?

Kannan171 (commented):
LGTM

PR #308, opened 2025-01-30
[brvndonkoh] iP

brvndonkoh:
# Clarawr Chatbot 🤖  

Clarawr is a simple chatbot that helps users **manage their tasks efficiently**. Users can add three types of tasks:  
- **ToDos**  
- **Deadlines**  
- **Events**  

Clarawr also allows users to **mark tasks as done** ✅ and keep track of their progress.  

## Features  
1. Add tasks (`todo`, `deadline`, `event`)  
2. Mark tasks as **done**  
3. View all tasks in a list  
4. Save and load tasks automatically  

## Usage Example  
Here’s an example of adding a **deadline** task in Clarawr:  
```java
Clarawr clarawr = new Clarawr();
clarawr.addTask("deadline submit report /by 2025-02-10");

### CODE REVIEW

File: src/main/java/Clarawr/Clarawr.java [hunk 1]
```diff
@@ -0,0 +1,257 @@
+package Clarawr;
```
weiyanminoo: Your main class currently has alot of responsibilities so you can consider separate certain parts of your code to different classes (ill add comments at the respective sections but up to u haha i think this is fine too)

src/main/java/Clarawr/Clarawr.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+import java.util.Scanner;
+
+enum TaskType {
```
weiyanminoo: can consider separating this part into an enum

src/main/java/Clarawr/Clarawr.java
```diff
@@ -0,0 +1,257 @@
...
+        scanner.close();
+    }
+
+    /**
+     * Handles the user input command by delegating to the appropriate method
+     * based on the parsed command.
+     *
+     * @param instruction The user input command.
+     * @throws ClarawrException If an error occurs while processing the command.
+     */
```
weiyanminoo: logic wise this is fine but for better readability you can consider placing this into the parser class

src/main/java/Clarawr/Clarawr.java
```diff
@@ -0,0 +1,257 @@
...
+			ui.showMessage("No matching tasks found.");
+		}
+	}
+
+    /**
+     * Adds a new Todo task to the task list.
+     *
+     * @param description The description of the Todo task.
+     * @throws ClarawrException If the description is empty.
+     */
```
weiyanminoo: i think it might be better if you add this and the subsequent methods into the TaskList class as they have the same responsibility of task management

src/main/java/Clarawr/Todo.java
```diff
@@ -0,0 +1,40 @@
...
+ * for the task type and its related behaviors.
+ */
+public class Todo extends Task {
+
+    /**
+     * Constructs a new Todo task.
+     *
+     * @param description A brief description of the task.
+     * @param isDone A boolean indicating whether the task is completed.
+     */
```
weiyanminoo: the constructor directly modifies isDone which breaks the encapsulation from the Task class so you cna consider calling markAsDone() instead

src/main/java/Level_1/Clarawr.java
```diff
@@ -0,0 +1,20 @@
+package Level_1;
+import java.util.Scanner;
+
```
weiyanminoo, line 3: i think this part is slightly confusing cos even though it is in a different package, you named this as Clarawr class too

weiyanminoo (commented):
hii brandon! overall good job on your code thus far! i just added some comments that could improve on the overall flow and readability but its not neccessary to change since your code logic is still correct. keep up the good work :))

src/main/java/Clarawr/Storage.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.IOException;
+
+import java.util.ArrayList;
+
+/**
+ * Handles saving and loading tasks to and from a file.
+ * The file path is hardcoded, and tasks are stored in a text file.
+ */
+public class Storage {
+    private static final String FILE_PATH = "C:\\Users\\user\\CS2103_IP\\ip\\src\\data\\clarawr.txt";
```
Wailydest: The file path is windows-specific, and it could fail when tried on MacOS. Maybe it would be better to use a system-independent way of achieving navigation?

src/main/java/Clarawr/Clarawr.java
```diff
@@ -0,0 +1,257 @@
...
+     * based on the parsed command.
+     *
+     * @param instruction The user input command.
+     * @throws ClarawrException If an error occurs while processing the command.
+     */
+    private static void handleCommand(String instruction) throws ClarawrException {
+        String[] commandParts = parser.parseCommand(instruction);
+        String command = commandParts[0].toLowerCase();
+
+		switch (command) {
```
Wailydest: According to the coding [standards](https://se-education.org/guides/conventions/java/intermediate.html) given to us, the spacing within the switch() statement should be 4-spaces long. It seems that this code use tabs instead. Maybe you could consider changing your editor formatting settings?

src/test/java/Clarawr/TaskListTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TaskListTest {
+    @Test
+    public void checkGetAllTask() {
```
Wailydest: Could tests be more comprehensive here? For example, the Storage class could be tested for it's ability to save and load from the storage

src/main/java/Clarawr/Storage.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Loads tasks from a file.
+     * Each line in the file is parsed into a task object.
+     * If corrupt data is encountered, the line is skipped.
+     *
+     * @return A list of tasks loaded from the file.
+     */
+    public ArrayList<Task> loadTasksFromFile() {
```
Wailydest: It's not quite clear what's the distinction between the member method `saveTasksToFile()` and the class method `loadTasksFromFile()`. Why one of them is static and the other is not, if the class don't have any member variables anyway? Maybe for consistency they both should have the same structure?

src/main/java/Clarawr/Storage.java
Wailydest: Also, it could be nice to have a way to pass a file path into Storage to make it more flexible instead of using hard-coded path. For example, when testing a Storage we'd want to make it save to a special testing directory instead of the default data one.

Wailydest (commented):
Good job! The structure of the project is quite transparent and it's easy to navigate over it. Few changes to it could still be made, including those specified by the previous reviewer. Also more testing units could definitely benefit the app. Overall, LGTM!

PR #318, opened 2025-01-30
[shashwat] iP

shashwatchan:
# Noga

> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://gettingthingsdone.com))

Noga frees your mind of having to remember things you need to do. It's:

- **text-based**
- **easy to learn**

All you need to do is:

1. Download it from [here](#).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😊.

And it's **FREE**!

## Features:
- ✅ Managing tasks  
- ✅ Managing deadlines 
- ⬜ Reminders *(coming soon)*  

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/noga/Event.java [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Represents a task that occurs during a specific time period.
+ * Extends the basic Task class with start and end times.
+ */
+public class Event extends Task {
+    private static final DateTimeFormatter INPUT_FORMAT = 
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HHmm");
+    private static final DateTimeFormatter OUTPUT_FORMAT = 
+            DateTimeFormatter.ofPattern("MMM d yyyy, h:mm a");
```
Billy026, line 14: Perhaps you could extract these into the Task class as protected variables? This way you can use them without having to duplicate it for every subclass of Task?

src/test/java/noga/DeadlineTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.*;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeParseException;
+
+public class DeadlineTest {
+    
+    @Test
+    public void testDeadlineCreationAndRetrieval() {
```
Billy026, line 11: Perhaps you should use this format for test functions: unitBeingTested_descriptionOfTestInputs_expectedOutcome, as described in the lecture topics?

src/main/java/noga/Noga.java
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * A task management chatbot that helps users keep track of various tasks,
+ * including todos, deadlines, and events.
+ */
+public class Noga {
+    private static final String DATA_DIR = "data";
+    private static final String DATA_FILE = "data/tasks.txt";
+    private Task[] tasks;
+    private int cur_index;
```
Billy026, line 23: Perhaps you should replace the variable name to curIndex instead?

src/main/java/noga/Noga.java
```diff
@@ -0,0 +1,387 @@
...
+            if (!found) {
+                System.out.println("No tasks found on this date.");
+            }
+        } catch (DateTimeParseException e) {
+            System.out.println("Please use the format yyyy-MM-dd (e.g., 2024-03-15)");
+        }
+    }
+
+    private void findTasks(String keyword) {
+        System.out.println("____________________________________________________________");
```
Billy026: Perhaps you can extract this into its own function? This way you can effectively abbreviate this.

Billy026 (commented):
The code looks all good, it complies with almost all JavaDoc guidelines. Personally, I feel that Noga.java is very long, which might make things harder to modify in future tasks.

src/main/java/noga/Noga.java
```diff
@@ -0,0 +1,387 @@
...
+                    System.out.println("Now you have " + (cur_index-1) + " tasks in the list.");
+                    System.out.println("____________________________________________________________");
+                    saveTasks();
+                } catch (NumberFormatException e) {
+                    System.out.println("____________________________________________________________");
+                    System.out.println("Please provide a valid task number!");
+                    System.out.println("____________________________________________________________");
+                }
+                continue;
+            }
```
mhaikelll, line 250: There seems to be a bit of code repetition when handling the various tasks. Perhaps you can extract this logic into separate helper methods or even a separate class to make it more maintainable.

src/main/java/noga/Noga.java
mhaikelll, line 23: Variable name should be in camelCase, perhaps switch to curIndex.

src/main/java/noga/Noga.java
```diff
@@ -0,0 +1,387 @@
...
+                try {
+                    String[] parts = userInput.split(" ");
+                    if (parts.length < 2) {
+                        System.out.println("____________________________________________________________");
+                        System.out.println("Please specify which task to mark (e.g., mark 1)");
+                        System.out.println("____________________________________________________________");
+                        continue;
+                    }
+                    int index = Integer.parseInt(parts[1]);
+                    if (index < 1 || index >= cur_index) {
```
mhaikelll: Perhaps can throw an exception or handle the error in a more graceful manner through a custom "InvalidTaskIndexException" as opposed to just printing a message.

src/main/java/noga/Noga.java
```diff
@@ -0,0 +1,387 @@
...
+                            task.isDone() ? 1 : 0,
+                            task.getDescription());
+                }
+                writer.write(line);
+                writer.newLine();
+            }
+            writer.close();
+        } catch (IOException e) {
+            System.out.println("Warning: Could not save tasks to file");
+        }
```
mhaikelll, line 124: Perhaps you can catch more specific exceptions in addition to IOException such as FileNotFoundException to provide additional context.

mhaikelll (commented):
Overall, the code looks good and errors are being handled generally well. The UI also adds a welcoming perspective to the chatbot, which contributes greatly to user experience. However, Noga.java could perhaps be separated into various classes such as Parser, TaskList and UI to potentially assist in making it more scalable in the future. Good job!

PR #325, opened 2025-01-31
[mhaikelll] iP

mhaikelll:
# Huhuhuharis

> “Your mind is for having ideas, not holding them.” – David Allen

Huhuhuharis frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use 

All you need to do is,

1. download it from [here](https://github.com/mhaikelll/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😗 

And it is **FREE!**

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

```
public static void main(String[] args) throws HuhuhuharisException {
        new Huhuhuharis("./data/huhuhuharis.txt").run();
}

public void run() {
        ui.showWelcomeMessage();
        while (true) {
            String input = ui.getUserInput();
            String reply = chatResponse(input);
            ui.showTaskList(reply);
            if (input.equals("bye")) {
                break;
            }
        }
        Storage.saveTasks(taskList.getTasks());
}

### CODE REVIEW

File: src/main/java/Huhuhuharis/Huhuhuharis.java [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+        if (description.isEmpty()) {
+            throw new HuhuhuharisException("Empty ToDo Description!");
+        }
+        taskList.addTask(new Todo(description));
+        Storage.saveTasks(taskList.getTasks());
+        return "Got it. I've added this task:\n" + taskList.getTask(taskList.getListCount() - 1) + "\nNow you have " + taskList.getListCount() + " tasks in the list.";
+    }
+
+    public static String deleteTask(String input) {
+        int taskId = Integer.parseInt(input.split(" ")[1]) - 1;
```
WhiteBear82: It looks like there isn’t a validation check for taskId before accessing taskList. If the input isn’t a valid number or is out of bounds, it might cause the program to crash. It might be helpful to add a check to ensure taskId is within a valid range before attempting to remove the task. That way, users won’t run into unexpected errors.

src/main/java/Huhuhuharis/TaskList.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public void unmarkTask(int taskId) {
+        taskLists.get(taskId).unmark();
+    }
+
+    public void addAll(ArrayList<Task> tasks) {
+        if (tasks == null) {
+            throw new NullPointerException("List cannot be null.");
+        }
+        tasks.addAll(taskLists);
```
WhiteBear82: Correct me if I'm wrong, but the logic here seems a bit reversed. Perhaps taskLists.addAll(tasks) would work better to ensure tasks loaded from storage are added correctly to the existing list?

src/main/java/Huhuhuharis/TaskList.java
```diff
@@ -0,0 +1,64 @@
+package Huhuhuharis;
+import java.util.ArrayList;
+
+public class TaskList {
+
+    private ArrayList<Task> taskLists;
```
WhiteBear82: The variable name taskLists seems a bit misleading since it holds a single list of tasks, not multiple lists. Would it be clearer to rename it to tasks? Just a suggestion for clarity!

src/main/java/Huhuhuharis/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.FileNotFoundException;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+
+    private String path;
```
WhiteBear82: It seems this variable is never used in the class. If it's unnecessary, would it be better to remove it for cleaner code?

src/main/java/Huhuhuharis/TaskList.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public ArrayList<Task> getTasks() {
+        return taskLists;
+    }
+    public int getListCount() {
+        return listCount;
+    }
+
+    public String fullList() {
```
WhiteBear82: Just a suggestion, but using StringBuilder instead of string concatenation inside a loop could improve performance. It might be worth considering for efficiency!

```java
StringBuilder fullList = new StringBuilder("Here are the tasks in your list:\n");
for (int i = 0; i < listCount; i++) {
    fullList.append(i + 1).append(".").append(taskLists.get(i)).append("\n");
}
return fullList.toString();
```

src/main/java/Huhuhuharis/Huhuhuharis.java
```diff
@@ -0,0 +1,119 @@
...
+
+public class Huhuhuharis {
+    private Ui ui;
+    private static TaskList taskList;
+    private Storage storage;
+
+    public Huhuhuharis(String path) throws HuhuhuharisException {
+        ui = new Ui();
+        storage = new Storage("./data/huhuhuharis.txt");
+        taskList = new TaskList(Storage.loadTasks());
```
WhiteBear82: It looks like tasks are being loaded twice. Would it be possible to load the tasks once to avoid duplication? Something like:
```java
taskList = new TaskList(Storage.loadTasks());
```

src/main/java/Huhuhuharis/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+            File file = new File("./data");
+            if (!file.exists()) {
+                file.mkdir();
+            }
+            FileWriter writer = new FileWriter(FILE_PATH);
+            for (Task task : tasks) {
+                writer.write(task.saveToFile() + "\n");
+            }
+            writer.close();
+        } catch (IOException e) {
```
WhiteBear82: It might be helpful to catch more specific exceptions in addition to IOException and provide additional context in the error messages. For example, you could separate FileNotFoundException and IOException and give a bit more detail on the error. It could look something like this:
```java
try {
    // file operations
} catch (FileNotFoundException e) {
    System.out.println("Error: The specified file was not found.");
} catch (IOException e) {
    System.out.println("Error: An unexpected I/O error occurred.");
}
```

WhiteBear82 (commented):
Your attention to detail is evident throughout the project. For example, the handling of specific task types with their own classes not only keeps the code clean but also makes it scalable for future features. The exception handling is robust, and I noticed that you're mindful of giving clear error messages, which enhances debugging and user experience. You've also ensured that the UI is welcoming and intuitive, which contributes positively to the overall user experience by making interactions feel smooth and straightforward. Keep up the great work!

src/main/java/Huhuhuharis/Parser.java
```diff
@@ -0,0 +1,29 @@
...
+public class Parser {
+
+    public static String dateTimeToStr(LocalDateTime dateTime) {
+        return dateTime.format(DateTimeFormatter.ofPattern("MMM dd yyyy hh:mm a"));
+    }
+
+    public static LocalDateTime strToDateTime(String str) throws DateTimeParseException {
+        try {
+            str = str.replaceAll("am", "AM").replaceAll("pm", "PM");
+            System.out.println("Trying to parse date-time: " + str);
```
Jasonzhou97: Looks good, maybe can reflect in GUI instead of printing?

src/main/java/Huhuhuharis/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+            if (!file.exists()) {
+                System.out.println("No task data found.");
+                return tasks;
+            }
+            Scanner sc = new Scanner(file);
+            while (sc.hasNextLine()) {
+                String line = sc.nextLine();
+                String[] parts = line.split(" \\| ");
+                Task task = null;
+                switch (parts[0]) {
```
Jasonzhou97: I think switch and case should be at same indentation level?

src/main/java/Huhuhuharis/Huhuhuharis.java
```diff
@@ -0,0 +1,139 @@
...
+                result.append(i + 1).append(".").append(task).append("\n");
+            }
+        }
+        if (!isFound) {
+            result.append("No tasks found matching the keyword: ").append(keyword);
+        }
+        return result.toString();
+    }
+
+    public static String markTask(String input) {
```
Jasonzhou97: Perhaps can add an exception handling here that handles if input length is less than 2?

Jasonzhou97 (commented):
Looks good overall!

PR #334, opened 2025-01-31
Jonathen Cheng Yuzhe ip

jonaturn:
BotJohn frees your mind of having to remember your todos. It's
- text-based
- easy to pick up
- ~~FAST~~ SUPER FAST

All you need to do is,

1. download it from [here](https://github.com/jonaturn/ip/tree/master)
2. double click it
3. add your task
4. let John manage your deadlines 😄 

At no cost:

Features: 

- [x] Manage Tasks
- [ ] Manage Deadlines (coming soon frfr)
- [ ] Reminders (coming soon frfr)

If you are a Java programmer, you can add your own features too. Start with the `main` method and branch out as you go.

```
public static void main(String[] args) throws Exception {
        // init
        new John().run();
    }
```


src/main/java/John.java
```diff
@@ -0,0 +1,53 @@
...
+        this.tasklist = new TaskList();
+    }
+
+    /**
+     * Initialises and runs the chatbot
+     *
+     * @throws IOException
+     */
+    public void run() throws IOException {
+        Scanner sc = new Scanner(System.in);
```
vrisdng, line 31: Since we are not taking user's input anymore, but loading it from a file and saving it to a file, where did you put this in your code?

src/main/java/parser/InputType.java
```diff
@@ -0,0 +1,43 @@
+package parser;
+
+/**
+ * Enums for commands
+ */
+public enum InputType {
```
vrisdng, line 6: Nice way to categorizes the input types here, but maybe you can extract it a little bit more? e.g: INVALID is not really an input type but should be an exception

src/main/java/tasks/Todo.java
```diff
@@ -0,0 +1,10 @@
+package tasks;
+
+/**
+ * Class that handles todo tasks
+ */
+public class Todo extends Task {
```
vrisdng, line 6: It feels quite weird to me that there are no methods in the subclasses of task, perhaps you omitted it? Maybe it is your program's choice of design, but I would recommend you to have methods that are specific to this class, e.g: An override toString method to describe Todo tasks.

vrisdng (commented):
Overall, you had a clear structure of code. There are some parts that have rooms for improvements. On additional note, your repository's name should be in the form of [your name] iP, else they might not detect yours correctly?

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Returns the name of the deadline and the formatted date and time
+     *
+     * @return description of deadline
+     */
+    @Override
+    public String name() {
+        return super.name() + " (by: " + by.format(DateTimeFormatter.ofPattern("dd MMM yyyy h:mma")) + ")";
+    }
```
Charlesl12: I think the method name could be improved here! Perhaps getName() or getDescription?

Charlesl12 (commented):
Great JavaDoc with complying conventions. Although some methods could follow the get...() naming conventions. LGTM.

PR #387, opened 2025-02-01
[J4Joshua] iP

J4Joshua:
# Ekko Chatbot Application

Ekko keeps track of your tasks. 🎉 :

Forked from [Github/CS2103T](https://github.com/nus-cs2103-AY2425S2/ip)

## Features
- Task management
- Command parsing
- User-friendly interface
- Persistent storage
- Error handling

## Key Components
1. Ekko Class - The main application entry point.
2. Storage Class - Handles data persistence.
3. Task Package - Manages the task classes and TaskList class
4. Ui Class - Manages user interactions.
5. Parser Class - Interprets user commands.
6. Command Package - Manages command classes
> [!TIP]
> See *command Package* for list of possible commands.

Here's the main method:
```java
        new Ekko("data/ekko.txt").run();
    }
```

Todo: 
- [x] Apply GMFD
- [ ] A-CheckStyle
- [ ] Level-10
- [ ] A-Varargs


src/main/java/Ekko.java
```diff
@@ -0,0 +1,49 @@
...
+            tasks = new TaskList(storage.load());
+        } catch (EkkoException e) {
+            ui.showLoadingError(e.getMessage());
+            tasks = new TaskList();
+        }
+    }
+
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
```
zedonggg: Perhaps it would be more descriptive to name the boolean isExitCommand?

src/main/java/command/Command.java
```diff
@@ -0,0 +1,30 @@
+package command;
+
+import task.TaskList;
+import ui.Ui;
+import storage.Storage;
+import exception.EkkoException;
+
```
zedonggg: Perhaps it would be neater to organize the import order alphabetically

src/main/java/command/Command.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Represents an executable command.
+ */
+public abstract class Command {
```
zedonggg, line 10: I like the use of OOP, especially for all the commands 👍

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws EkkoException {
+        Task removedTask = tasks.delete(index);
+
+        ui.showLine();
+        ui.showMessage("Noted. I've removed this task:");
+        ui.showMessage("[" + removedTask.getCategory() + "][" + removedTask.getStatusIcon() + "] " + removedTask);
+        ui.showMessage("Now you have " + tasks.size() + " tasks in the list.");
```
zedonggg: Would it be better to combine it as one String? That way, you only need to call showMessage once

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Parses user input and returns the corresponding Command object.
+ */
+public class Parser {
+    public static Command parse(String input) throws EkkoException {
+        String[] inputArr = input.split(" ", 2);
+        String commandWord = inputArr[0];
+
+        switch (commandWord) {
+            case "list":
```
zedonggg: Double check just in case but I believe the course style guide recommends the case statements to be inline with the switch statement

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,102 @@
+package storage;
+
+import exception.EkkoException;
+import task.Deadline;
+import task.Event;
+import task.Task;
+import task.ToDo;
+
+import java.io.*;
```
zedonggg: Would it be good for the specific libraries to be specified rather than a wildcard import?

zedonggg (commented):
LGTM. Love the good usage of OOP. Your code is clear and very pleasant to read. Just some minor nits here and there. 👍

PR #390, opened 2025-02-01
[Li Yaner] iP

liy722:
# DukePro

> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://example.com))
DukePro free your mind of having to remember things you need to do. It's,

- textbased
- easy to learn
- ~~**FAST**~~ **SUPER FAST** to use

All you need to do is,

1. download it from [here](https://example.com).
2. double-click it.
3. add your tasks.
  
And it is *FREE*

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

___
```java
public class Main {
       }
}

### CODE REVIEW

File: src/main/java/Todo.java [hunk 1]
```diff
@@ -0,0 +1,10 @@
+public class Todo extends Task{
```
itsapaulblem: Missing whitespace before the {

src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+class DukeException extends Exception {
```
itsapaulblem: class should be named AException as chatbot is named as A

src/main/java/A.java
```diff
@@ -0,0 +1,94 @@
...
+            try {
+                String input = scanner.nextLine();
+
+                if (input.equals("bye")) {
+                    System.out.println("Bye. Hope to see you again soon!");
+                    break;
+                }
+
+                if (input.equals("list")) {
+                    if(count==0){
```
itsapaulblem: should have a whitespace before the {

src/main/java/A.java
```diff
@@ -0,0 +1,94 @@
...
+                    System.out.println("Now you have " + count + " tasks in the list.");
+                } else if (input.startsWith("event ")) {
+                    String description = input.substring(6).split(" /from ")[0];
+                    String[] time = input.substring(6).split(" /from ")[1].split(" /to ");
+                    list.add(new Event(description, time[0], time[1]));
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println(" " + list.get(count));
+                    count++;
+                    System.out.println("Now you have " + count + " tasks in the list.");
+                }else if (input.startsWith("delete ")) {
```
itsapaulblem: Missing whitespace before the else if statement

src/main/java/A.java
```diff
@@ -0,0 +1,94 @@
...
+                    if (index < 0 || index >= count) {
+                        throw new DukeException("Task number out of range.");
+                    } else {
+                        Task removedTask = list.remove(index);
+                        count--;
+                        System.out.println("Noted. I've removed this task:\n" + " "+removedTask);
+                        System.out.println("Now you have " + count + " tasks in the list.");
+                    }
+                }
+                else{
```
itsapaulblem: Missing whitespace after the else statement

src/main/java/A.java
```diff
@@ -0,0 +1,94 @@
...
+                    count++;
+                    System.out.println("Now you have " + count + " tasks in the list.");
+                }else if (input.startsWith("delete ")) {
+                    int index = Integer.parseInt(input.substring(7)) - 1;
+                    if (index < 0 || index >= count) {
+                        throw new DukeException("Task number out of range.");
+                    } else {
+                        Task removedTask = list.remove(index);
+                        count--;
+                        System.out.println("Noted. I've removed this task:\n" + " "+removedTask);
```
itsapaulblem: missing whitespace before the variable removedTask

src/main/java/A.java
```diff
@@ -0,0 +1,94 @@
...
+                        System.out.println("Now you have " + count + " tasks in the list.");
+                    }
+                }
+                else{
+                    throw new DukeException("please enter a valid task");
+                }
+            }catch (DukeException e) {
```
itsapaulblem: Missing whitespace before the catch statement

itsapaulblem (commented):
LGTM. Just some cosmetic changes.

src/main/java/A.java
GaaneshT: For this, I think you can create a parser class and handle it there. You can also use a switch statement to make it more readable.

src/main/java/A.java
GaaneshT: Instead of count, I would suggest list.size(), it's more robust

src/main/java/Deadline.java
```diff
@@ -0,0 +1,15 @@
+public class Deadline extends Task {
+
+    protected String by;
```
GaaneshT: I think it's better to make by private, since it does not allow direct modification from child classes and does not violate encapsulation

GaaneshT (commented):
LGTM, a suggestion I can make would be to follow a more OOP-based approach, it would make it cleaner to read! Overall, good work so far! :)

PR #391, opened 2025-02-01
[Liu Jihan] iP

Liu-Jihan:


src/main/java/OscarL.java
```diff
@@ -80,6 +81,7 @@ public static void main(String[] args) {
                         tasks.add(newDeadline);
```
Windofbitter: I like the way you organize the code, it is really clean with the trailing comment. May be we can try to separate the add task logic unit with the ui output logic by a few line to make it more organized?

src/main/java/OscarL.java
```diff
@@ -94,6 +96,22 @@ public static void main(String[] args) {
                         tasks.add(newEvent);
                         System.out.println("Got it. I've added this task:");
                         System.out.println(newEvent);
+                        System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                        break;
+
+                    case "delete": // Remove a task
+                        if (inputParts.length < 2) {
```
Windofbitter: It is good that you follow coding standard strictly by adding four space for statements like conditions, it is a common mistake.

src/main/java/Task.java
```diff
@@ -21,4 +24,46 @@ public void markAsNotDone() {
...
+
+        if (parts.length < 3) {
+            throw new IllegalArgumentException("Invalid task format in file.");
+        }
+
+        String type = parts[0];
+        boolean isDone = parts[1].equals("1");
+        String description = parts[2];
+
+        switch (type) {
```
Windofbitter: Maybe we can try to align with the coding standard that switch and case should have the same indentation?

src/main/java/Task.java
```diff
@@ -21,4 +24,46 @@ public void markAsNotDone() {
...
+                ToDo todo = new ToDo(description);
+                if (isDone) todo.markAsDone();
+                return todo;
+
+            case "D":
+                if (parts.length < 4) {
+                    throw new IllegalArgumentException("Invalid deadline format in file.");
+                }
+                Deadline deadline = new Deadline(description, parts[3]);
+                if (isDone) deadline.markAsDone();
```
Windofbitter: Maybe we can also wrap the single statement with brackets as well to follow the coding standard?

src/main/java/Command.java
```diff
@@ -0,0 +1,110 @@
...
+    public boolean isExit() {
+        return isExit;
+    }
+
+    public static Command parse(String input, TaskList tasks, Ui ui, Storage storage) throws OscarLException {
+        String[] parts = input.split(" ", 2);
+        String command = parts[0];
+
+        switch (command) {
+            case "bye":
```
Casielim: IIRC, there is no need to indent the case clauses according to Java coding standards. You might want to adjust that.

src/main/java/Command.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public void execute() {
+        action.run();
+    }
+
+    public boolean isExit() {
+        return isExit;
+    }
+
+    public static Command parse(String input, TaskList tasks, Ui ui, Storage storage) throws OscarLException {
```
Casielim: Try to avoid long methods, as they often contain more information, making debugging harder later. Consider shortening them if possible.

Casielim (commented):
Looks good to me.
Here are some suggestions that won't affect the functionality of your code.

src/main/java/oscarl/OscarL.java
```diff
@@ -0,0 +1,63 @@
...
+import ui.Ui;
+import task.TaskList;
+import java.util.ArrayList;  // Missing import for ArrayList
+/**
+ * The main class for the OscarL task management application.
+ * Handles initialization, user interaction, and command execution.
+ */
+public class OscarL {
+    private Storage storage;
+    private TaskList tasks;
```
varuuuun: The `TaskList` object you are referrencing to here is a single object, not the list itself - hence it should be named as `task` or `taskList` instead of `tasks`?

src/main/java/task/Task.java
```diff
@@ -0,0 +1,118 @@
...
+            throw new IllegalArgumentException("Invalid task format in file.");
+        }
+
+        String type = parts[0];
+        boolean isDone = parts[1].equals("1");
+        String description = parts[2];
+
+        switch (type) {
+            case TODO_TYPE:
+                ToDo todo = new ToDo(description);
```
varuuuun, line 86: Inconsistent naming? if class name is `ToDo`, then shouldn't object name be `toDo`?

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    /**
+     * Returns the number of tasks in the list.
+     *
+     * @return The size of the task list.
+     */
+
+    public int size() {
```
varuuuun, line 94: Should be named as `getSize()`. Use verb instead of noun.

src/main/java/command/Command.java
```diff
@@ -0,0 +1,174 @@
...
+     * Parses a user input string and returns the corresponding Command.
+     *
+     * @param input The user input string.
+     * @param tasks The TaskList containing the current tasks.
+     * @param ui The UI component for displaying messages.
+     * @param storage The Storage component for saving task data.
+     * @return The parsed Command object.
+     * @throws OscarLException If the input command is invalid.
+     */
+    public static Command parse(String input, TaskList tasks, Ui ui, Storage storage) throws OscarLException {
```
varuuuun: Method name might be a bit generic, you might want to consider `parseStringToCommand` instead?

varuuuun (commented):
Some nits on naming.

PR #396, opened 2025-02-01
[Charlesl12] iP

Charlesl12:
# Clovis

> "Your mind is for having ideas, not holding them." - David Allen [source](https://dansilvestre.com/productivity-quotes)

Clovis helps you to remember everything you need to do. It's
- text-based
- easy to learn
- ~~FAST~~ **SUPER FAST** to use

All you have to do is,

1. download it from [here](https://github.com/Charlesl12/ip).
2. add your tasks
3. let it manage your tasks for you 👍 

And it is **FREE**!

Features:
- [X] Managing tasks
- [X] Managing deadlines
- [X] Managing events

```java  
public class Main {
        Application.lauch(MainApp.class, args);
    }
}

### CODE REVIEW

File: src/main/java/Clovis.java [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+import java.util.ArrayList;
+import java.time.LocalDateTime;
+
+public class Clovis {
+    public static void main(String[] args) throws IOException {
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+        ArrayList<Task> list = Storage.loadTasks();
+        System.out.println("Hello! I'm Clovis.\nWhat can I do for you?");
+
+        enum Command {
```
Bryce-3D: Nice use of enums for the commands

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
...
+                    case "bye" -> BYE;
+                    default -> UNKNOWN;
+                };
+            }
+        }
+
+        while(true) {
+            try {
+                String input = br.readLine();
+                String[] split_input = input.split(" ", 2);
```
Bryce-3D: Hi! I think this variable should be renamed as something like `splitInput` since the codestyle says that in general, variables should be written in `camelCase`. It also says that collections of stuff like lists should be in plural form, but given the variable here, I'm not really sure if `splitInputs` is exactly accurate either.

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
...
+                Command command = Command.fromString(split_input[0]);
+
+                switch(command) {
```
Bryce-3D: Hi! The codestyle guide for CS2103T mentions that the switch case block should not have an extra indent for the case clauses, so it should look something like this
```java
switch (cmd) {
case LIST:
    // Something
    break;
case BYE:
    // Something
    // ...
```

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
...
+                                System.out.println((i + 1) + ". " + list.get(i));
+                            }
+                        }
+                        break;
+                    case BYE:
+                        Storage.saveTasks(list);
+                        System.out.println("Bye. Hope to see you again soon!");
+                        return;
+                    case MARK:
+                        int mark_index = Integer.parseInt(split_input[1]);
```
Bryce-3D: Hi! I think this is another variable that does not follow the `camelCase` naming convention, and I see quite a few more below. I think it might be a good idea to ctrl+F for `_`s in your code to find and rename variables from `snake_case` to `camelCase`. I also made this mistake at first and had to spend a bit of time fixing it a week or two ago.

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
...
+                        } else if (taskType.equalsIgnoreCase("deadline")) {
+                            String[] splitDescription = description.split("/by ", 2);
+                            if (description.isEmpty() || splitDescription.length < 2) {
+                                throw new ClovisException("Unacceptable, a description and /by date are required for a deadline");
+                            }
+                            list.add(new Deadline(splitDescription[0], splitDescription[1]));
+                        } else if (taskType.equalsIgnoreCase("event")) {
+                            String[] splitDescription = description.split("/from | /to ");
+                            if (description.isEmpty() || splitDescription.length < 3) {
+                                throw new ClovisException("Unacceptable, a description, /from and /to date are required for an event");
```
Bryce-3D: Hi! I think this line, when including the indent, exceeds 120 characters. In the CS2103T style guide, it mentions 110 chars as a soft limit and 120 chars as a hard limit for the line length, so maybe a line break can be added somewhere.

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
...
+                        String description = split_input.length > 1 ? split_input[1] : "";
+                        if (taskType.equalsIgnoreCase("todo")) {
+                            if (description.isEmpty()) {
+                                throw new ClovisException("Unacceptable, a description is required for a todo");
+                            }
+                            list.add(new ToDo(description));
+                        } else if (taskType.equalsIgnoreCase("deadline")) {
+                            String[] splitDescription = description.split("/by ", 2);
+                            if (description.isEmpty() || splitDescription.length < 2) {
+                                throw new ClovisException("Unacceptable, a description and /by date are required for a deadline");
```
Bryce-3D: Hi! I think this line, like the one below, likely exceeds the 120 chars hard limit for line length

src/main/java/Event.java
```diff
@@ -0,0 +1,36 @@
...
+        this.end = LocalDateTime.parse(endDateTime, OUTPUT_FORMAT);
+    }
+
+    public String toFileFormat() {
+        return "D | " + (isDone ? "1" : "0") + " | " + description + " | " + start.format(OUTPUT_FORMAT) + " | " + end.format(OUTPUT_FORMAT);
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + "(from: " + start.format(OUTPUT_FORMAT) + " to: " + end.format(OUTPUT_FORMAT) + ")";
```
Bryce-3D: Hi! I think that this line also exceeds 120 chars, so a line break would be needed here too.

Bryce-3D (commented):
Hi! Another general comment is that there is currently no javadocs available. As listed under the [`A-JavaDocs` task from week 3](https://nus-cs2103-ay2425s2.github.io/website/schedule/week3/project.html#5-add-increments-as-parallel-branches-a-javadoc-a-codingstandard-level-9)
> Minimal: Add header comments to at least half of the non-private classes/methods.

I hope this review helps!

src/main/java/Clovis.java
```diff
@@ -0,0 +1,109 @@
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.time.LocalDateTime;
+
+public class Clovis {
+    public static void main(String[] args) throws IOException {
```
TasmiaH0508: Perhaps it would be better to group chunks of code and turn them into methods so that readability of the code is improved in general.

src/main/java/Event.java
```diff
@@ -0,0 +1,36 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Event extends Task {
```
TasmiaH0508: I think it would be good to have JavaDoc comments for as many methods and classes as possible.

TasmiaH0508 (commented):
Generally, there are very few errors in following the java coding standard. 

Code quality, in the naming aspect, is good as descriptive names were given to variables and functions. 

However, there is a lack of JavaDoc comments.

PR #403, opened 2025-02-01
[blacksesamezongzi] iP

blacksesamezongzi:
AdventureGuide helps you out in this exciting yet sometimes upsetting adventure of life, by recording all the things you need to do at one step. It's,
- text-based
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

Steps to start your journey with AdventureGuide...
1. download jar file from github
2. double click the jar file
3. add your tasks
4. let it manage your tasks for you 😺 

And it is **FREE**!

Features at hand:
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/components/AdventureGuideBot.java
```diff
@@ -0,0 +1,154 @@
+package components;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+import Exceptions.*;
```
YosieSYX: Would it be better to expand out explicitly each package that you imported here? So then the users will be able to know exactly what packages they need.

src/main/java/components/AdventureGuideBot.java
```diff
@@ -0,0 +1,154 @@
...
+        tasks.add(new Deadline(description, by));
+        System.out.println("____________________________________________________________");
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + tasks.get(tasks.size() - 1));
+        System.out.println(" Now you have " + tasks.size() + " tasks in the list.");
+        System.out.println("____________________________________________________________");
+    }
+
+    private void handleEvent(String input) throws EmptyDescriptionException {
+        String[] parts = input.substring(6).split(" /from | /to ");
```
YosieSYX: Would it be better to give this variable 'parts' a more meaningful name? Because currently it's not too clear for me what exactly this string stores.

src/main/java/components/AdventureGuideBot.java
```diff
@@ -0,0 +1,154 @@
...
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + tasks.get(tasks.size() - 1));
+        System.out.println(" Now you have " + tasks.size() + " tasks in the list.");
+        System.out.println("____________________________________________________________");
+    }
+
+    private void handleDelete(String input) throws InvalidTaskNumberException {
+        int taskIndex = Integer.parseInt(input.split(" ")[1]) - 1;
+        if (taskIndex >= 0 && taskIndex < tasks.size()) {
+            Task removedTask = tasks.remove(taskIndex);
```
YosieSYX: I personally feel that the variable name 'removedTask is a bit misleading because it is actually the task that is about to be removed but not yet removed. Would a name like: requestedToBeRemovedTask be better?

src/main/java/components/Task.java
```diff
@@ -0,0 +1,28 @@
+package components;
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
YosieSYX: I like this boolean variable name.

YosieSYX (commented):
Overall good and clean code. Some minor issues I identified are that some variable names are not meaningful enough. Consider adjusting a bit if that makes sense.

src/main/java/components/AdventureGuideBot.java
ypuppy: would it fits the coding style better if you dont use wildcard importing ?

src/main/java/components/AdventureGuideBot.java
```diff
@@ -0,0 +1,154 @@
...
+    private void handleDeadline(String input) throws EmptyDescriptionException {
+        String[] parts = input.substring(9).split(" /by ");
+        String description = parts[0].trim();
+        String by = parts[1].trim();
+        if (description.isEmpty() || by.isEmpty()) {
+            throw new EmptyDescriptionException("deadline");
+        }
+        tasks.add(new Deadline(description, by));
+        System.out.println("____________________________________________________________");
+        System.out.println(" Got it. I've added this task:");
```
ypuppy: would it be better if you separate the print function and input taking function in other files so it cleaner and easier for debugging and make changes later on

ypuppy (commented):
I like your code, its very neat.

PR #410, opened 2025-02-01
[Kkenjji] iP

Kkenjji:
# 👾  Clank: Chatbot x Task Manager

“A plan is a list of actions arranged in whatever sequence is thought likely to achieve an objective.” – **John Argenti**

Clank is a intelligent **task management chatbot** that helps you stay organized and productive. 🧠  

## ✨ Features  
- **Text-based** interaction 📝  
- **Easy to learn** 📖  
- **Blazing fast** ⚡  
- **Completely FREE!** 🎉  

## 🛠 How to Use  
1. **Download** the latest version from [here](https://github.com/kkenjji/ip/releases).  
2. **Double-click** to run the application or run it using the command:
```
java -jar clank.jar
```
3. **Add your tasks** via simple text commands.  
4. Let Clank do the rest! 🤖  

## ✅ Upcoming Features  
- [x] Basic task management  
- [x] GUI support  
- [x] Reminder notifications 

## 🔧 Code Example  
Here's how a task is added by **Clank**:  

```java
public void addTask(Task task) {
    tasks.add(task);
}
```

---

🚀 **Stay productive with Clank!**


src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+            String second = (splitted.length == 2) ? splitted[1] : "";
+
+            System.out.println(line);
+
+            try {
+                switch (first) {
+                case "bye":
+                    System.out.println(byeMessage);
+                    System.exit(0);
+                    break;
```
nicholasohjj: Avoid Using `System.exit(0)` in a Loop-Based Application

It forces an abrupt termination instead of allowing a graceful exit.

```suggestion
                case "bye":
                    System.out.println(byeMessage);
                    break; // Breaks out of the while loop, terminating normally
```

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+
+        String byeMessage = " Bye. Hope to see you again soon!\n";
+        System.out.println(welcomeMessage);
+
+        Scanner scanner = new Scanner(System.in);
+        String input;
+
+        while (true) {
+            input = scanner.nextLine();
+            input = input.toLowerCase();
```
nicholasohjj: Avoid `toLowerCase()` on Entire Input

This can cause unintended issues when parsing user commands that involve case-sensitive data.

```suggestion
            input = scanner.nextLine();
            String[] splitted = input.split(" ", 2);
            String first = splitted[0].toLowerCase(); // Convert only the command
            String second = (splitted.length == 2) ? splitted[1] : "";
```

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+        String line = "____________________________________________________________\n";
+        String welcomeMessage = " Greetings! I'm Clank, Ratchet's best friend.\n" +
+                " I heard you are in trouble, and I'm here to help.\n" +
+                " How may I assist you?\n" +
+                line;
+
+        String byeMessage = " Bye. Hope to see you again soon!\n";
+        System.out.println(welcomeMessage);
+
+        Scanner scanner = new Scanner(System.in);
```
nicholasohjj: The `Scanner `object is never closed, leading to a potential resource leak.

Use a `try-with-resources` statement to ensure `scanner` is automatically closed.

```suggestion
        try (Scanner scanner = new Scanner(System.in)) {
    while (true) {
        String input = scanner.nextLine();
        // (rest of the loop remains unchanged)
    }
}
```

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+            input = input.toLowerCase();
+
+            String[] splitted = input.split(" ", 2);
+            String first = splitted[0];
+            String second = (splitted.length == 2) ? splitted[1] : "";
+
+            System.out.println(line);
+
+            try {
+                switch (first) {
```
nicholasohjj: Improve Readability of `switch` Statements

The `switch` cases handle different user commands but some cases call methods directly, while others throw exceptions. This inconsistency makes it harder to read.

Use a helper method like `processCommand()` to separate command handling logic from input reading.

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+                case "event":
+                    taskManager.addEvent(second);
+                    break;
+                default:
+                    throw new UnknownCommandException(first);
+                }
+            } catch (ClankException e) {
+                System.out.println(e.getMessage());
+            } catch (Exception e) {
+                System.out.println("An unexpected error occured: " + e.getMessage());
```
nicholasohjj: Improve Exception Handling for Unexpected Errors

The catch block for general exceptions prints only the error message, which might not be useful for debugging.

```suggestion
                System.out.println("An unexpected error occured: " + e.getMessage());
                e.printStackTrace(); // Prints the full stack trace

```

nicholasohjj (changes requested):
nits but they enhance code structure, maintainability, and reliability.

LGTM after these refinements! 👍

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.ArrayList;
+
+public class Clank {
+    private static TaskManager taskManager = new TaskManager();
+
+    public static void main(String[] args) {
+        String line = "____________________________________________________________\n";
+        String welcomeMessage = " Greetings! I'm Clank, Ratchet's best friend.\n" +
+                " I heard you are in trouble, and I'm here to help.\n" +
+                " How may I assist you?\n" +
```
ShiJianXi: Might have some minor coding standards violation here, refer to the [coding standard guide](https://se-education.org/guides/conventions/java/intermediate.html), on the section on Layout.
Should have breaks before an operators.

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+                line;
```
ShiJianXi:
```suggestion
        String welcomeMessage = " Greetings! I'm Clank, Ratchet's best friend.\n" 
                     + " I heard you are in trouble, and I'm here to help.\n" 
                     + " How may I assist you?\n" 
                     + line;
```

src/main/java/Clank.java
```diff
@@ -0,0 +1,69 @@
...
+        Scanner scanner = new Scanner(System.in);
+        String input;
+
+        while (true) {
+            input = scanner.nextLine();
+            input = input.toLowerCase();
+
+            String[] splitted = input.split(" ", 2);
+            String first = splitted[0];
+            String second = (splitted.length == 2) ? splitted[1] : "";
```
ShiJianXi: Perhaps a more meaningful variable name can be used instead of first and second.

ShiJianXi (commented):
LGTM. Code is clean, just some nits on coding standards.

src/main/java/clank/command/AddCommand.java
```diff
@@ -0,0 +1,73 @@
...
+import clank.utility.Storage;
+
+public class AddCommand extends Command {
+    private final Task task;
+
+    public AddCommand(String input) {
+        String[] parts = input.split(" ", 2);
+
+        String mainCommand = parts[0].toLowerCase();
+        String[] details = parts.length > 1 ? parts[1].split("/by | /from | /to ", 3) : new String[]{""};
```
shelialia: Could this variable name be more specific?

src/main/java/clank/command/AddCommand.java
```diff
@@ -0,0 +1,73 @@
...
+            System.out.println("It is from " + start + " to " + end + ".");
+            System.out.println("Check it out!");
+            break;
+        default:
+            throw new ClankException("Unknown task type.");
+        }
+    }
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) {
```
shelialia, line 90: Could this method name be more specific? "Execute" is rather vague, a reader may find difficulty understanding the purpose of the method. Perhaps we could rename it to a specific action such as "addAndStoreTask"

src/main/java/clank/task/Task.java
```diff
@@ -0,0 +1,84 @@
...
+    public static Task fromSavedFormat(String line) {
+        String[] parts = line.split("\\|");
+
+        if (parts.length < 3) {
+            System.out.println("Corrupted task.");
+            return null;
+        }
+
+        try {
+            String type = parts[0];
```
shelialia: "type" is rather vague as well, and is frequently used in coding concepts. perhaps we could use "taskType" instead?

src/main/java/clank/task/Task.java
```diff
@@ -0,0 +1,84 @@
...
+            boolean isDone = Boolean.parseBoolean(parts[1]);
+            String description = parts[2];
+
+            Task task = null;
+            switch (type) {
```
shelialia: Good job on following the coding standard indentation for the switch-case statement!

src/main/java/clank/task/Task.java
```diff
@@ -0,0 +1,84 @@
+package clank.task;
+
+public abstract class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
shelialia, line 17: Good job on ensuring the boolean sounds like a boolean.

shelialia (commented):
LGTM. Several occurrences of vagueness in variable names, but overall clean code 👍

src/main/java/clank/command/AddCommand.java
Kkenjji: Oh, this method overrides the execute method in its parent class (Command.java).

src/main/java/clank/task/Task.java
Kkenjji: Thanks for the recommendation. Will modify accordingly.

src/main/java/clank/command/AddCommand.java
Kkenjji: I'll change it to taskDetails, more specific.

src/main/java/Clank.java
Kkenjji: Thanks for the advice. I have changed the variable names.

src/main/java/Clank.java
Kkenjji: I see, I've forgotten about that. Thanks for the reminder!

src/main/java/Clank.java
Kkenjji: Will change accordingly, thanks!

src/main/java/Clank.java
Kkenjji: I have extracted all different commands out as classes of their own, thanks for the advice!

src/main/java/Clank.java
Kkenjji: Fixed!

src/main/java/Clank.java
Kkenjji: Fixed!

src/main/java/Clank.java
Kkenjji: Fixed!

src/main/java/Clank.java
Kkenjji: Thanks for the reminder!

PR #411, opened 2025-02-01
[pranavyey] iP

pranavyey:
### yeyAI

> "The true art of memory is the art of attention." - Samuel Johnson 

yeyAI helps you remember things you need to do. It's,

- text-based
- easy to learn
- **SUPER FAST** to use
- All you need to do is,

1. download it from [here](https://github.com/pranavyey/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

let it manage your tasks for you 😉
And it is FREE!

Features:

- [x]  Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)

```java
        new YeyAI(System.getProperty("user.dir") + "/tasks.txt").run();
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,16 @@
+public class Deadline extends Task{
```
Benjam11n: Do ensure that there is a space between Task and { in the method signature

src/main/java/yeyAI.java
```diff
@@ -0,0 +1,153 @@
...
+
+public class yeyAI {
+    public static void main(String[] args) {
+        ArrayList<Task> tasks = new ArrayList<>();
+        String home = System.getProperty("user.home");
+        java.nio.file.Path path = java.nio.file.Paths.get(home, "Documents", "2103T", "ip", "tasks.txt");
+        if (Files.exists(path)) {
+            try {
+                List<String> lines = Files.readAllLines(path);
+                for (String taskLine : lines){
```
Benjam11n: Do also ensure that there is a space between for (String taskLine : lines) and { in the for loop.

src/main/java/yeyAI.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        String line = " ____________________________\n";
+        Scanner scanner = new Scanner(System.in);
+
+
+        System.out.println(line + "Hello! I'm yeyAI\n" + "What can I do for you?\n" + line);
+        String input = scanner.nextLine();
+        while (!input.equals("bye")) {
+            try {
```
Benjam11n: Would it be better to place `String input = scanner.nextLine();` within the while loop? This would eliminate the need for repeated input = scanner.nextLine(); statements in each branch and prevent potential infinite loops if someone forgets to read the next input.

src/main/java/yeyAI.java
```diff
@@ -0,0 +1,153 @@
...
+                            System.out.printf("%d.%s\n", displayIndex, tasks.get(i).toString());
+                        }
+                    }
+                    input = scanner.nextLine();
+                } else if (input.startsWith("mark")) { //mark branch
+                    int taskIndex = Integer.parseInt(input.split(" ")[1]) - 1;
+                    tasks.get(taskIndex).setDone();
+                    System.out.println("Nice! I've marked this task as done:");
+                    System.out.println(tasks.get(taskIndex).toString());
+                    input = scanner.nextLine();
```
Benjam11n: I noticed that for the deadline branch, you checked for whether the task the user intended to delete was actually in the list. Could similar checks for the unmark and mark commands be implemented?

```
if (Integer.parseInt(parts[1]) > tasks.size()) {
    throw new YeyException("Task to be deleted must exist within list");
}
```

Benjam11n (commented):
Overall good work so far!

PR #427, opened 2025-02-02
[Maliha Haque] iP

malihahaque:
### Lara

>"For every minute spent organizing, an hour is earned." – Benjamin Franklin

Lara is a smart task manager that helps you stay organized by keeping track of your tasks efficiently.

**Features**
- Text-based interface📝 
-  Easy to use 🚀
-  Fast and efficient⚡

**How to Use**

1. Download the application from [here](https://github.com/malihahaque/ip.git)📥 
2. Double-click to open
3. Add your tasks and manage them easily✍ 
4.  Let Lara handle the rest for you!🎯

**Upcoming Features**

 - [x] Managing tasks

 - [ ] Reminders (coming soon)

**Code Example**
_Here's the main method to launch the application:_

```java
public class Main {
        Application.launch(Lara.class, args);
    }
}

### CODE REVIEW

File: src/main/java/Lara.java [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+        } catch (LaraException e) {
+            System.out.println(e.getMessage());
+        }
+    }
+
+    private static void printList() {
+        System.out.println("Here are the tasks in your list:");
+        for (int i=1; i<=list.size(); i++) {
```
lesterlimjj: Your operators (e.g. "=", "<=", "-") should be surrounded by space on both sides. This minor issues reappear a few times in other parts of your code(e.g. addTask(String) method).

src/main/java/Lara.java
```diff
@@ -0,0 +1,122 @@
...
+                    break;
+                } else if (message.equals("list")) {
+                    printList();
+                } else if (firstWord.equals("mark")) {
+                    markTask(message);
+                } else if (firstWord.equals("unmark")) {
+                    unmarkTask(message);
+                } else if (firstWord.equals("delete")) {
+                  deleteTask(message);
+                } else if (firstWord.equals("todo") || firstWord.equals("deadline") || firstWord.equals("event")){
```
lesterlimjj: Perhaps we can split into more if-else statements and make each task creation its own function?

src/main/java/LaraException.java
```diff
@@ -0,0 +1,5 @@
+public class LaraException extends Exception{
```
lesterlimjj: { should have a white space preceding it.

src/main/java/Task.java
```diff
@@ -0,0 +1,34 @@
+import java.util.ArrayList;
+
+public class Task {
```
lesterlimjj: Consider putting your object classes into packages (naming for packages need to be all lowercase). It is a week 3 IP requirement but it would be good to start doing it in earlier branches.

src/main/java/Task.java
```diff
@@ -0,0 +1,34 @@
...
+    protected String description;
+    protected boolean isDone;
+    protected static int taskCounter=0;
```
lesterlimjj: Perhaps we can remove this if we do not plan to use it?

src/main/java/Lara.java
```diff
@@ -0,0 +1,122 @@
...
+                list.add(deadline);
+                System.out.println(deadline);
+            } else {
+                String[] parts= date.split(" /from | /to ", 3);
+                Task event = new Event(parts[0], parts[1], parts[2]);
+                list.add(event);
+                System.out.println(event);
+            }
+            System.out.println("Now you have " + list.size() + " tasks in the list.");
+        } catch (LaraException e) {
```
lesterlimjj: Do we need a try-catch for every task-related function call if we already have a try-catch in your if-else statements?

lesterlimjj (commented):
You are making good progress in catching up. Aside from a few minor coding violations and some long pieces of code, everything else look fine. Before you progress into later weeks of the IP, it would be good to install the checkstyle scripts from week 4 to help you check coding standard violation. Aside from that, configuring your IDE to fit the coding standard is available in week 2 or 3 so perhaps you can skim through and configure it now.

src/main/java/Lara.java
```diff
@@ -0,0 +1,122 @@
...
+        try {
+            int taskNumber = Integer.parseInt(command.split(" ")[1]) - 1;
+            if (taskNumber < 0 || taskNumber >  list.size()) {
+                throw new LaraException("Please input a valid task number!");
+            }
+                list.get(taskNumber).markAsUndone();
+                System.out.println("OK, I've marked this task as not done yet:");
+                System.out.println("[ ] " + list.get(taskNumber).getDescription());
+            } catch (LaraException e) {
+            System.out.println(e.getMessage());
```
kylieluk88: This line should have another 4-space indentation. some of the spacing for the "}" brackets below also seem a bit inconsistent, can make use of IDEs which automatically adjust the indentation for you so you don't have to manually check :)

kylieluk88 (commented):
other than some small issues like spacing and indexing, the code is clear and easy to understand! keep it up in the week 3 tasks

PR #428, opened 2025-02-02
[Siyan] iP

Siyan-G:
# Chatty 🗨️  
> “Productivity is never an accident. It is always the result of a commitment to excellence, intelligent planning, and focused effort.” – Paul J. Meyer  

Meet **Chatty**, your personal task management assistant! Unlike traditional to-do apps, Chatty interacts with you just like a friendly assistant, helping you keep track of your tasks with ease. It's:  

- **Conversational** – Manage tasks naturally through chat 💬  
- _Easy to use_ – No complicated commands needed! 🎓  
- ~~Slow~~ **Lightning fast** – No more forgetting deadlines ⚡  

## Getting Started 🚀  
Using Chatty is as simple as:  

1. **Download** it from [here](#).  
2. **Run it** with a double-click.  
3. **Chat with Chatty** to manage your tasks.  
4. Stay on top of your schedule effortlessly! 😉  

## Features ✨  
- [x] Create and track **todos**  
- [x] Schedule and manage **events**  
- [x] Set **deadlines** with ease  

## For Developers 👨‍💻👩‍💻  
Chatty is built in Java, making it a great project for Java developers to explore object-oriented programming. Here’s a look at the `main` method:  

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Chatty.java [hunk 1]
```diff
@@ -0,0 +1,176 @@
...
+import exception.ChattyInvalidArgumentException;
+import exception.ChattyInvalidInputException;
+import exception.ChattyTaskNotFoundException;
+
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Chatty {
+    private static final String linebreak = "___________________________________________________________________";
+    private static final String introMsg = "\n\tHello Master! I'm Chatty, your ever-ready personal assistant.\n\tHow can I help you today?";
```
kumar2215: I like how you created some constants here for some of the bots' messages. However, note that according to the Java Coding standard in this course, constants use the screaming snake case instead of the camel case used for normal variables.

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+    private static final String exitMsg = "\n\tGoodbye Master! See you soon!";
+    private ArrayList<Task> taskList = new ArrayList<>();
+    private Integer taskCounter = 0;
```
kumar2215: Any reason why you used the Wrapper class `Integer` instead of the primitive type `int`?

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    public void Exit() {
+        this.Response(exitMsg);
+    }
+
+    public void handleCommand(String command) {
+        try{
+            // handle list command
+            if (command.equalsIgnoreCase("list")) {
```
kumar2215: The other commands do not allow for case insensitivity. So why is the list command special? Maybe consider changing command to a single case (lower or upper) first before processing it if you want to allow for case insensitivity that is.

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+        } catch (ChattyException e) {
+            this.Response(e.getMessage());
+        }
+
+    }
+
+    public void deleteTask(String command) throws ChattyTaskNotFoundException, ChattyInvalidArgumentException {
+        String[] commandParts = command.split(" ");
+        if (commandParts.length == 2) {
+            int taskId = Integer.parseInt(commandParts[1].trim());
```
kumar2215: Perhaps consider the case where the user types "delete study" where they didn't think of using the index at first. This would cause an `NumberFormatException` when trying to parse the 2nd word as a number which isn't handled.

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+                response.append(String.format("\n\t%d. %s", i + 1, taskList.get(i)));
+            }
+            this.Response(response.toString());
+        }
+    }
+
+    public void setTaskStatus(String command) throws ChattyTaskNotFoundException {
+        String[] commandParts = command.split(" ");
+        if (commandParts.length == 2) {
+            int taskId = Integer.parseInt(commandParts[1].trim());
```
kumar2215: Same issue as the one mentioned for `deleteTask`. Perhaps consider placing the parsing of the index number in its own method since its being used in more than 1 place.

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+            } else {
+                taskList.get(taskId - 1).unmarkDone();
+                response.append("\n\tOkay! I've marked this task as not done yet:\n");
+                response.append(String.format("\t%s", taskList.get(taskId - 1)));
+            }
+            this.Response(response.toString());
+        }
+    }
+
+    public void handleTaskCreation(String command) throws ChattyInvalidArgumentException {
```
kumar2215: Perhaps the processing of the description string for each task can be handled in the respective Task subclasses instead?

src/main/java/Chatty.java
```diff
@@ -0,0 +1,176 @@
...
+            taskList.add(new Todo(taskName));
+            taskCounter++;
+            response.append("\n\tI've added this task to the list:\n");
+            response.append(String.format("\t%s\n", taskList.get(taskCounter - 1)));
+            response.append(String.format("\tNow you have %d tasks in the list.", taskCounter));
+            this.Response(response.toString());
+        }
+        if (command.startsWith("event")) {
+            String taskString = command.substring(5).trim();
+            String[] taskParts = taskString.split("/");
```
kumar2215: What is the allowed format for event tasks here? If it is "name /from start /to end", the strings "from" and "to" are not checked, meaning the string "name /start start /end end" might be incorrectly parsed.

src/main/java/exception/ChattyInvalidArgumentException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private final String commandType;
+
+    public ChattyInvalidArgumentException(String commandType) {
+        this.commandType = commandType;
+    }
+
+    @Override
+    public String getMessage() {
+        return String.format(" Invalid argument detected for command: %s", commandType);
```
kumar2215: Maybe informing the user of what the correct format should be can be useful.

src/main/java/Chatty.java
kumar2215, line 1: A lot of the methods used in the class seem to be called internally in this file alone. Perhaps consider making those that are not going to be called outside this class to be made private?

kumar2215 (commented):
Here are some issues that came to my mind. Certain issues might be a bit nitpicking while the rest are more concerned with user functionality and something you might want to consider looking into.

src/main/java/Chatty.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public void Intro() {
+        System.out.println(LINE_BREAK);
+        this.Response(INTRO_MSG);
+    }
+
+    public void Exit() {
+        this.Response(EXIT_MSG);
+    }
```
themintchoco: Method names should be in camelCase

src/main/java/Deadlines.java
themintchoco, line 1: It may be better to name this class `Deadline` (singular form), since it represents a single deadline.

src/main/java/Events.java
themintchoco, line 1: Similar to `Deadlines`, it may be better to name this class `Event` (singular form), since it represents a single event.

themintchoco (commented):
Thanks for the PR! Just a few minor nits. Great work!

src/main/java/Chatty.java
Siyan-G: thanks for the input, have made the relevant changes!

src/main/java/Chatty.java
Siyan-G: thanks for the input, have changed it to primitive type

src/main/java/Chatty.java
Siyan-G: thanks for the input, have made relevant changes to make it consistent

src/main/java/Chatty.java
Siyan-G: thanks for the input, it is really helpful, will make relevant changes

src/main/java/Chatty.java
Siyan-G: thanks for the input, it is really helpful, will make relevant changes

src/main/java/exception/ChattyInvalidArgumentException.java
Siyan-G: thanks for the input, will make relevant changes

src/main/java/Chatty.java
Siyan-G: thanks for the input, will make relevant changes

src/main/java/Chatty.java
Siyan-G: thanks for the input, will make relevant changes

src/main/java/Chatty.java
Siyan-G: thanks for the input, will make relevant changes

src/main/java/Chatty.java
Siyan-G: thanks for the input, will make relevant changes

src/main/java/Deadlines.java
Siyan-G: thanks for the input, have made the relevant changes

src/main/java/Events.java
Siyan-G: thanks for the input, have made the relevant changes

PR #429, opened 2025-02-02
[Keith Tan] iP

nathtiekk:
# Kif

> Kif frees your mind of having to remember things you need to do.

It is:
- text-based
- fast to use
- ~easy~ SUPER EASY to learn

All you need to do is:
1. download it from [here.](https://github.com/nathtiekk/ip/releases/tag/A-Gradle)
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😃 

Features:
- [X] Managing tasks
- [ ] Managing deadlines (_coming soon_)
- [ ] Reminders (_coming soon_)

> [!TIP]
> If you are a Java programmer, you can use it to practice Java too. Here's the `main` method:

```java
public class Main {
    }
}
```


src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class Kif {
+
+    private static void introduce() {
+        String introduce =
+                """
```
changjy81, line 9: Huh, I wasn't aware of text blocks in Java. Good idea to use them!
Though I'm not sure of the code style requirements for their indentation...

src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
...
+                        System.out.println(e.getMessage());
+                    }
+                }
+                default -> System.out.println(
+                                    """
+                                    ____________________________________________________________
+                                    OOPS!!! I'm sorry, but I don't know what that means :-(
+                                    ____________________________________________________________"""
+                );
+            }
```
changjy81, line 68: The case statements shouldn't be indented according to the coding standard?

src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
```
changjy81, line 4: I see that the week 3 iP tasks were not pushed yet, but the classes should be in packages and should have header Javadoc comments on classes/methods

src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
...
+    private static boolean isTerminate(String code) {
+        return code.equals("bye");
+    }
+
+    public static void main(String[] args) throws IOException {
+        BufferedReader reader = new BufferedReader(new InputStreamReader(System.in));
+        introduce();
+        String userMessage = reader.readLine();
+        String[] splitMessage = userMessage.split(" ");
+        while(!isTerminate(splitMessage[0])) {
```
changjy81, line 46: Shouldn't there be a whitespace after while here?
Noticed similar occurences in the rest of the code as well.

changjy81 (changes requested):
put wrong review type; redone below

src/main/java/Kif.java
changjy81: I see that the week 3 iP tasks were not pushed yet, but the classes should be in packages and should have header Javadoc comments on classes/methods

src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
...
+public class Kif {
+
+    private static void introduce() {
+        String introduce =
+                """
+                ____________________________________________________________
+                Hello! I'm Kif
+                What can I do for you?
+                ____________________________________________________________
+                """;
```
changjy81: Huh, I wasn't aware of text blocks in Java. Good idea to use them!
Though I'm not sure of the code style requirements for their indentation...

src/main/java/Kif.java
changjy81: Shouldn't there be a whitespace after while here?
Noticed similar occurences in the rest of the code as well.

src/main/java/Kif.java
```diff
@@ -0,0 +1,74 @@
...
+                    } catch (KifException e) {
+                        System.out.println(e.getMessage());
+                    }
+                }
+                default -> System.out.println(
+                                    """
+                                    ____________________________________________________________
+                                    OOPS!!! I'm sorry, but I don't know what that means :-(
+                                    ____________________________________________________________"""
+                );
```
changjy81: The case statements shouldn't be indented according to the coding standard?

changjy81 (commented):
Outside of some recurring coding standard violations, LGTM overall.

src/main/java/Task.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+class Task {
+    private static final ArrayList<Task> userTasks = new ArrayList<>();
+    protected String description;
+    protected boolean isDone;
+    private static final String FILEPATH = "tasks.txt";
+    protected taskType type;
+    private static final String KEYWORD = "kifReservedKeyword";
```
gohlucas: Would grouping variables by their access modifiers be better? For instance, grouping the private static final variables together etc...

src/main/java/Task.java
```diff
@@ -0,0 +1,271 @@
...
+                    Task t = new Event(description[1], description[2], description[3]);
+                    t.isDone = Boolean.parseBoolean(description[0].trim());
+                }
+                else if (description.length == 3) {
+                    Task t = new Deadline(description[1], description[2]);
+                    t.isDone = Boolean.parseBoolean(description[0].trim());
+                }
+            }
+        } catch (FileNotFoundException e) {
+            //do nothing
```
gohlucas: I would recommend handling the exception. This could be in the form of printing an error message to the user to inform them that there was an issue with the ongoing process

src/main/java/Task.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    public static void printTotalTasks() {
+        System.out.println("Now you have " + Task.userTasks.size() + " tasks in the list.");
+    }
+
+    private static void editTaskTXT (int lineNumber, Kif.UserCommand operation) {
+        ArrayList<String> preItems = new ArrayList<>();
+        ArrayList<String> postItems = new ArrayList<>();
+        ArrayList<String> editLine = new ArrayList<>();
```
gohlucas: Would changing the variable name to a plural form be more inline with the Java Coding Standard for collection of objects?

src/main/java/Task.java
```diff
@@ -0,0 +1,271 @@
...
+                }
+            }
+            fw.close();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void listUserTask() {
+        System.out.println("____________________________________________________________");
```
gohlucas: Would extracting the line into a static final variable provide a cleaner body for the methods? I noticed this line is used quite extensively in this file

src/main/java/Task.java
```diff
@@ -0,0 +1,271 @@
...
+            System.out.println(Integer.toString(counter) + "." + task);
+            counter++;
+        }
+        System.out.println("____________________________________________________________");
+    }
+
+    public static void markUserTask(int key) {
+        Task currentTask = userTasks.get(key - 1);
+        currentTask.isDone = true;
+        editTaskTXT(key, Kif.UserCommand.MARK);
```
gohlucas:
```suggestion
        editTaskTxt(key, Kif.UserCommand.MARK);
```

gohlucas: Perhaps this would be more in line with the java coding standard

src/main/java/Kif.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public enum UserCommand {
+        LIST,
+        MARK,
+        UNMARK,
+        DEADLINE,
+        EVENT,
+        TODO,
+        DELETE;
+    }
```
gohlucas: I love that you created an enum for the types of commands!!

gohlucas (commented):
I can see the vision. Looking forward to Kif's development 👍

PR #439, opened 2025-02-02
[Arnav Bajpai] iP

arnavbajpai:
# Krypto ChatBot ⚓

Krypto is a command-line chatbot that helps you manage your tasks efficiently. It supports adding, deleting, marking, and searching for tasks.  

## Features  
- Add tasks: ToDos, Deadlines, and Events  
- Delete tasks  
- Mark tasks as done/undone  
- List all tasks  
- Search for tasks by keyword  
- Save and load tasks from a file  

## Krypto is
* text-based
* easy to learn
* ~~FAST~~ **SUPER FAST** to use

All you need to do is,
1. download it from [here](https://github.com/arnavbajpai/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😋

And it is **FREE**!

## Features:

- [x] Managing tasks
- [x] Managing deadlines (coming soon)
- [ ] Reminders (coming soon)
- [ ] Recording events and rescheduling

```java
        new Krypto(FILE_PATH).run();
    }

### CODE REVIEW

File: src/main/java/Task.java [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+        this.isDone = true;
+        this.sign = "X";
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println(this);
+    }
+    public void unmarkTask() {
+        this.isDone = false;
+        this.sign = "";
+        System.out.println("OK, I've marked this task as not done yet:");
+        System.out.println(this);
```
zedonggg: I like the use of OOP within your program. Do you think it would be better to return a String rather than print out? Your invalid and incomplete commands return Strings so perhaps it would be more organized to have everything return Strings to one method that can then separately print everything

zedonggg (commented):
LGTM. You're probably still in the process of doing the other iP tasks so all the best in getting those done! Some suggestions from me would be to split everything into packages, and further extract out all the methods into objects (such as different command types or Ui classes). And remember to add your javadoc comments! [👍]

src/main/java/Krypto.java
```diff
@@ -0,0 +1,107 @@
...
+                printResponseWithLines("Great talking to you!");
+                break;
+            }
+            else if (prompt.equals("list")) {
+                printList(arr);
+                continue;
+            }
+            else if (first.equals("mark")) {
+                int id = Integer.parseInt(split[1]);
+                arr.get(id-1).markTask();
```
matthanfoo: can consider adding white space before and after operators: operators should be surrounded by a space character.

src/main/java/Krypto.java
```diff
@@ -0,0 +1,107 @@
...
+                continue;
+            }
+            else if (first.equals("unmark")) {
+                int id = Integer.parseInt(split[1]);
+                arr.get(id-1).unmarkTask();
+                continue;
+            }
+            else if (first.equals("delete")) {
+                int id = Integer.parseInt(split[1]);
+                arr.remove(id -1);
```
matthanfoo: slightly inconsistent spacing here as well

src/main/java/Task.java
```diff
@@ -0,0 +1,41 @@
...
+        if (prompt.length <= 1) {
+            return "";
+        }
+        for(int i = 1; i < prompt.length; i ++) {
+            content.append(prompt[i]).append(" ");
+        }
+        return content.toString();
+    }
+
+    public void markTask() {
```
matthanfoo: agreed with @zedonggg, i think this breaks abstraction slightly as all prints should be executed by the Krypto bot rather than by the Task class itself

matthanfoo (commented):
really well organised code! just a few syntax change suggestions

arnavbajpai:
# Krypto ChatBot ⚓

> "Your mind is for having ideas, not holding them." - David Allen ([source](https://en.wikipedia.org/wiki/Dave_Allen_(comedian)))

DukePro frees your mind of having to remember things you need to do. It's,

* text-based
* easy to learn
* ~~FAST~~ **SUPER FAST** to use

All you need to do is,

1. download it from [here](https://github.com/arnavbajpai/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😋

And it is **FREE**!

## Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```java
        Application.launch(Krypto.class, args);
    }
}

PR #441, opened 2025-02-03
[Timothy Tay] iP

tim0tay:
Maybe this will be able to do that one day! But for now: 
Prophet frees your mind of having to remember things you need to do. It's:
- text-based
- easy to understand
- ~FAST~ SUPER FAST to use

All you need to do is: 
1. download it from [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ&pp=ygUIcmlja3JvbGw%3D)
2. double-click it.
3. add your tasks.

And it is **free**!

Features: 
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Managing events
- [ ] Reminders (coming soon)
---

If you are a Java programmer, you can use this to practice too. Here's the `main` method: 
```java
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,27 @@
+public class Deadline extends Task {
+    protected static final String type = "[D]";
```
awc1116, line 2: Would it be better to make this private instead of protected since it is not used by any other classes in the same package?

src/main/java/Prophet.java
```diff
@@ -0,0 +1,177 @@
+import java.util.Scanner;
+public class Prophet {
+    protected TodoList tasks;
+    private static final String horizontalLine = "--------------------------------------------------------------------------------\n";
+    private static final String usage = "Please type a proper format..!\n";
+    public Prophet() {
+        this.tasks = new TodoList();
+    }
+    public void inputInterpretator(String str) throws UnrecognisedCommandException {
```
awc1116, line 9: Would it be cleaner to create a separate class of inputInterpretator so that you do not cluster your main class?

src/main/java/Prophet.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+        } catch (ProphetException e) {
+            System.out.println(e.toString());
+        }
+    }
+
+    /**
+     * Prints a string that lists all tasks.
+     */
+    public void printList() {
```
awc1116, line 36: Same for all these methods. Would it be easier to create separate classes for the different techniques call so that when you want to add, edit, or delete certain commands or methods it would be easier and cleaner?

src/main/java/AddDeadlineCommand.java
```diff
@@ -0,0 +1,35 @@
+import java.time.LocalDate;
+
+public class AddDeadlineCommand extends Command {
+    /**
+     * The AddDeadlineCommand class represents the command to add a deadline task.
+     */
+    private static final String horizontalLine = "-------------------------------------------" +
```
soonami69: I like how you broke this long line so as to not make the line too long!

src/main/java/AddToDoCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Adds a to-do task to the list of tasks.
+     * Prints out the confirmation message or the error message.
+     * @param ui the user interface of the chatbot
+     * @param storage the storage of tasks accumulated as the chatbot runs
+     */
+    @Override
+    public void execute(Ui ui, Storage storage) {
+        ToDoTask newTask = new ToDoTask(this.taskDescription);
+        String added = storage.addToList(newTask);
```
soonami69: Perhaps a more descriptive name instead of `added` would be better?

src/main/java/Parser.java
```diff
@@ -0,0 +1,192 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+
+public class Parser {
+    /**
+     * The Parser class takes in user input and interprets it to perform the necessary actions.
+     */
+    private static final String horizontalLine = "--------------------------------------------------------------------------------\n";
```
soonami69: I think this line may be too long. Consider breaking it into two?

src/main/java/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+        // if first word is in commandtype, parse further for args
+        String firstWord = str.split(" ", 2)[0].toLowerCase();
+        ArrayList<Command> commands = new ArrayList<>();
+        switch (firstWord) {
+        case "list":
+            commands.add(new ListCommand(CommandType.LIST));
+            return commands;
+        case "mark":
+            try {
+                String[] mark = str.split("mark ", 2);
```
soonami69: Perhaps a more intuitive variable name here? I don't quite understand what `mark`'s purpose is based on its name.

src/main/java/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+                }
+                String[] timeline = remainingParts[1].trim().split("to: ",2);
+                if (timeline.length < 2 || timeline[1].trim().isEmpty()) {
+                    throw new NoDescriptionException();
+                }
+                LocalDate from = LocalDate.parse(timeline[0].trim(), DateTimeFormatter.ofPattern("MMM dd yyyy"));
+                LocalDate to = LocalDate.parse(timeline[1].trim(), DateTimeFormatter.ofPattern("MMM dd yyyy"));
+                commands.add(new AddEventCommand(
+                        CommandType.EVENT, remainingParts[0].trim(), from, to));
+                if (isDone) {
```
soonami69: I like how you named the boolean variables intuitively!

src/main/java/Ui.java
```diff
@@ -0,0 +1,61 @@
...
+        System.out.println(horizontalLine
+                + "Hi! Welcome to the Daily Prophet! You can call me Prophet for short.\n"
+                + "For now, I can keep track of your tasks. What would you like to remember?\n"
+                + horizontalLine);
+    }
+
+    /**
+     * Prints the farewell message of the chatbot.
+     */
+    public static void bye() {
```
soonami69: I think this should be a verb. Maybe `sayGoodbye`?

soonami69 (commented):
Nice, readable code, with good adherence to code style and quality guidelines 👍

PR #442, opened 2025-02-03
[Lee Miao En Emelyn] iP

potatoems:


src/main/java/Brownie.java
```diff
@@ -0,0 +1,102 @@
+import java.util.Scanner;
+
```
NhatMinh0208, line 2: Most of the public methods and classes are missing a Javadoc comment. Perhaps you can add comments to at least the classes?

src/main/java/Brownie.java
```diff
@@ -0,0 +1,102 @@
...
+public class Brownie{
```
NhatMinh0208, line 3: I think there should be a space before the brackets here?

src/main/java/Brownie.java
```diff
@@ -0,0 +1,102 @@
...
+                } else if (taskDescription.startsWith("event")) {
+                    if (scriptLength < 7) {
+                        throw new DukeException("Hello! Can I confirm what task you would like to note down?\n" +
+                                "Event tasks must look like: event xxxxxxxx /from xxx /to xxx");
+                    }
+                    tasks.add(taskDescription.substring(6), 'E');
+                    System.out.println(horizontalLine);
+                } else if (taskDescription.startsWith("deadline")) {
+                    if (scriptLength < 10) {
+                        throw new DukeException("Hello! Can I confirm what task you would like to note down?\n" +
```
NhatMinh0208, line 85: Can you push the `+` down to the following line?

src/main/java/Tasks.java
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            // in case user tries to mark/unmarked an already marked/unmarked item
+            System.out.println("It seems like it's already been done");
+        }
+    }
+
+    public void deleteTask(int taskNumber) {
+        Task task = tasks.get(taskNumber);
+        tasks.remove(taskNumber - 1);
+        System.out.println("Noted. I've removed this task:\n" + task + "\nNow you have " + tasks.size() + " tasks in the list.");
```
NhatMinh0208, line 61: This line is just a littttle bit too long. Can you split it into two?

src/main/java/Brownie.java
```diff
@@ -0,0 +1,102 @@
+import java.util.Scanner;
+
+public class Brownie{
+    private static final String horizontalLine = "____________________________________________________________";
```
NhatMinh0208, line 4: This is a constant, right? So I think the name should be HORIZONTAL_LINE?

src/main/java/Brownie.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        Tasks tasks = new Tasks();
+
+        String greeting = "\nHello! I'm BROWNIE\nWhat can I do for you?\n";
+        String farewell = "Bye. Hope to see you again soon!\n";
```
NhatMinh0208, line 11: And these strings are constants, too?

NhatMinh0208 (commented):
### 1st review
LGTM 👍🏻 Standard mostly followed, just need to clean up a few issues.

PR #445, opened 2025-02-03
# Chatty

matthanfoo:
# Chatty

Chatty helps you keep track of the million things on your to-do list, but it's also 
- text-based
- easy to learn
- SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/matthanfoo/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

Here's what an example task list looks like:

> Here are the tasks in your list:
> 1. [T] [ ] borrow book
> 2. [D] [X] return book /by 05-06-2022 (by: 2022-06-05T00:00)
> 3. [E] [ ] project meeting /from 12-02-2025 14:00 /to 12-02-2025 17:00 (from: 2025-02-12T14:00 to: 2025-02-12T14:00)

```
public class Chatty {
      new Chatty("ChattyData.csv").run();
   }
}
```
**TRY IT OUT NOW!**


src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+
+public class Chatty {
+    public static void main(String[] args) {
+
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+        System.out.println("Hello! I'm Chatty McChatface");
+        System.out.println("What can I do for you?");
+        ArrayList<Task> userInputs = new ArrayList<Task>();
+        try {
+            String s = br.readLine();
```
Nano-233: Should this have a more descriptive variable name? such as userInput

src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+            while (s != null) {
+                System.out.println("____________________________________________________________");
+                if (s.equals("bye") || s.equals("Bye")) {
```
Nano-233: Would using `s.equalsIgnoreCase("bye")` here be more efficient and readable?

src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+                } else if (s.equals("list")) {
+                    int i = 0;
+                    System.out.println("Here are the tasks in your list:");
+                    while (i != userInputs.size()) {
+                        System.out.println(String.valueOf(i + 1) + ". " + userInputs.get(i));
+                        i++;
+                    }
+                } else if (s.contains("mark")) {
+                    Pattern p = Pattern.compile("([0-9]+$)");
+                    Matcher m = p.matcher(s);
```
Nano-233: Any reason this two line definition is used multiple times instead of using a helper method?
(for example again in line 52-53)

src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+
+                        Deadline d = new Deadline(deadlineTitle, byString);
+                        userInputs.add(d);
+                        System.out.println("Got it. I've added this task:\n  " + d);
+                        System.out.println("Now you have " + userInputs.size() + " items in the list.");
+
+                    } else {
+                        System.out.println("Invalid item");
+                        break;
+                    }
```
Nano-233: Should switch-case be used instead of this many if-else statements to increase readability and maintainability?

src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+                                task.unmark();
+                                System.out.println("OK, I've marked this task as not done yet:");
+                                System.out.println("   " + task);
+                            } else if (s.contains("mark")) {
+                                task.mark();
+                                System.out.println("Nice! I've marked this task as done:");
+                                System.out.println("   " + task);
+                            }
+                        } else {
+                            System.out.println("Invalid item");
```
Nano-233: For error messages, should there be logics or other ways to implement them so that they are not  hardcoded?

src/main/java/Chatty.java
```diff
@@ -0,0 +1,152 @@
...
+                        }
+
+                        Deadline d = new Deadline(deadlineTitle, byString);
+                        userInputs.add(d);
+                        System.out.println("Got it. I've added this task:\n  " + d);
+                        System.out.println("Now you have " + userInputs.size() + " items in the list.");
+
+                    } else {
+                        System.out.println("Invalid item");
+                        break;
```
Nano-233: This may break the app as it breaks out of the while loop

src/main/java/Chatty.java
Nano-233, line 1: Should this code be refactored to improve readability and logic so that there are less hardcoding involved?

Nano-233 (commented):
I like how many potential user errors are considered and accounted for. Perhaps the code could be cleaner by refactoring?

PR #459, opened 2025-02-05
[Kea Harvan Suyanto] iP

kea-S:


src/main/java/shep/Shep.class
SomneelSaha2042, line 1: You should add .class files to your gitignore, otw there is uneccessary bloat.

src/main/java/shep/command/Commands.java
SomneelSaha2042, line 1: As compared to 1 Commands class, we can try to extract the functionality out into multiple classes.

src/main/java/shep/task/Deadline.java
SomneelSaha2042, line 1: when parsing the date the code should be put in a try catch block to make sure we handle incorrect input

src/main/java/shep/task/Event.java
SomneelSaha2042, line 1: try catch block should be used

SomneelSaha2042 (commented):
LGTM! 
A few small issues here and there but otw seems g!

src/main/java/Interaction.java
```diff
@@ -1,54 +1,28 @@
 import java.util.Scanner;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 
 public class Interaction {
     TaskList list;
```
bkkuan: Fields should generally be set to private for encapsulation

src/main/java/Interaction.java
```diff
@@ -1,54 +1,28 @@
...
-    Path filePath;
+    Storage storage;
 
     public Interaction() {
         this.list = new TaskList();
-
-        // load or initialise data directory
-        Path dataDirectoryPath = Paths.get("../data");
```
bkkuan: Its better to define constants for better maintainability
eg. private static final String DATA_DIRECTORY = "../data";
Then , Path dataDirectoryPath = Paths.get(DATA_DIRECTORY);

src/main/java/Interaction.java
```diff
@@ -1,54 +1,28 @@
...
 
     public void start() {
+        // load storage into tasklist
+        this.storage.writeInto(this.list);
+
         Scanner userInputScanner = new Scanner(System.in);
         boolean saidBye = false;
         while (!saidBye) {
             String currUserInputText = userInputScanner.nextLine();
-            saidBye = Commands.executeCommand(currUserInputText, this.list, this.filePath, true);
```
bkkuan: executeCommand() is called twice which may cause unintended behaviour.

src/main/java/TaskList.java
```diff
@@ -1,10 +1,7 @@
-import java.util.List;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.io.FileWriter;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.ArrayList;
```
bkkuan: java.util.ArrayList is imported twice. Remove the duplicate

src/main/java/TaskList.java
```diff
@@ -42,53 +39,4 @@ public String toString() {
...
-            }
-
-            index++;
-        }
-
-    }
-
-    public void writeToFile(Path filePath) {
-        // overwrite previous taskList to be empty, no need to add anything to the file
-        try (FileWriter fileWriter = new FileWriter(filePath.toString(), false)) {
```
bkkuan: this block is redundant
FileWriter is opened but not used, which is pointless

src/main/java/TaskList.java
```diff
@@ -1,10 +1,7 @@
-import java.util.List;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.io.FileWriter;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.ArrayList;
 
 public class TaskList extends ArrayList<Task> {
```
bkkuan: generally a bad practice as it violates encapsulation

bkkuan (commented):
Code is well structured. Variables and method names follow Java conventions. Well done!

PR #486, opened 2025-02-08
[Guo Fengming] ip

Gu0Fengming:


VikramGoyal23:
There seems to be several small coding standard violations, but beyond that I only have a couple of nitpicks.

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
+import java.io.BufferedReader;
```
VikramGoyal23: You should put this class in a package, maybe `main` or `whiost`.

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
...
+        String marked = "Nice! I've marked this task as done:";
+        String unmarked = "OK, I've marked this task as not done yet:";
+        String reportTask1 = "Now you have ";
+        String reportTask2 = " tasks in the list.";
+        String deleted = "Noted. I've removed this task:";
+
+        String error0 = "OOPS!!! I'm sorry, but I don't know what that means.";
+        String errorEmpty = "OOPS!!! The description cannot be empty.";
+        String errorNotFound = "OOPS!!! The task you select doesn't exist.";
+        String errorEmptyList = "OOPS!!! There's no task.";
```
VikramGoyal23: Since all of these strings are never reassigned, you should consider writing the variable names in SCREAMING_SNAKE_CASE to indicate they are constants.

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
...
+    public static void main(String[] args) {
+        String greeting = "Hello! I'm Whiost\nWhat can I do for you?\n";
+        String addTask = "Got it. I've added this task:";
+        String marked = "Nice! I've marked this task as done:";
+        String unmarked = "OK, I've marked this task as not done yet:";
+        String reportTask1 = "Now you have ";
+        String reportTask2 = " tasks in the list.";
+        String deleted = "Noted. I've removed this task:";
+
+        String error0 = "OOPS!!! I'm sorry, but I don't know what that means.";
```
VikramGoyal23: `error0` is a quite undescriptive variable name, maybe `errorUnknown` or `errorInvalidInput` could be better.

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
...
+        String errorEmpty = "OOPS!!! The description cannot be empty.";
+        String errorNotFound = "OOPS!!! The task you select doesn't exist.";
+        String errorEmptyList = "OOPS!!! There's no task.";
+
+        // Print greeting
+        System.out.print(greeting);
+
+        boolean p = true;
```
VikramGoyal23: `p` is a very undescriptive variable name, maybe something like `isInputting` could be better (it should at least sound like a boolean variable).

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
...
+        List<String> lst = new ArrayList<String>();
+        List<String> typeLst = new ArrayList<String>();
+        List<String> markLst = new ArrayList<String>();
+
+        try {
+            BufferedReader data = new BufferedReader(new FileReader("./data/whiost.txt"));
+            List<String> temp_lst = new ArrayList<String>();
```
VikramGoyal23: Variables should not be written in snake_case, you should change the variable name to `tempLst`.

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.PrintWriter;
+import java.io.IOException;
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Whiost {
+    public static void save(List<String> lst, List<String> typeLst, List<String> markLst) {
```
VikramGoyal23: Consider adding a JavaDoc header for this method to explain what it does..

src/main/java/Whiost.java
```diff
@@ -0,0 +1,168 @@
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.PrintWriter;
+import java.io.IOException;
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Whiost {
```
VikramGoyal23: Consider adding a JavaDoc header for the class to explain what it does.

src/main/java/Whiost.java
VikramGoyal23, line 1: Try and use more abstraction (e.g. separate classes) for the different types of tasks the bot will keep track of.

VikramGoyal23 (commented):
There seems to be several small coding standard violations, but beyond that I only have a couple of nitpicks.

